Issue #12, opened 2019-06-19
DALFs with missing dependencies crash the package parser


da-tanabe:
If `PackageService` returns DALFs where at least one DALF depends on a DALF that is unknown and not part of the same `PackageService` method, `dazl` crashes. It should instead emit a warning, and save the package for later processing.

PR #14, opened 2019-06-21
python: Crashing fix for incomplete dependencies

da-tanabe:
Prevent dazl from crashing when encountering incomplete DALF dependencies. Addresses #12 .


python/tests/unit/test_package_loading.py
```diff
@@ -0,0 +1,72 @@
...
+from operator import setitem
+from pathlib import Path
+from unittest import TestCase
+from zipfile import ZipFile
+
+from dazl import sandbox, Network
+from dazl.model.types_store import PackageStore
+from dazl.util.dar import build_dar, DarFile
+
+CACHE_DIR = Path(__file__).parent.parent.parent.parent / '.cache' / 'test-dars'
```
mschaef-da: Is there a better way to get this path? (Pretty please. :-) )

python/tests/unit/test_package_loading.py
```diff
@@ -0,0 +1,72 @@
...
+DAML_FILE: Path = Path(__file__).parent.parent / 'resources' / 'AllKindsOf.daml'
+
+
+class PackageLoadingTest(TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        # Intentionally create a DAR with missing dependencies
+        build_dar(DAML_FILE, DAR_FILE, allow_caching=True)
+        if DALF_FILE.exists() and DALF_FILE.lstat().st_mtime >= DAR_FILE.lstat().st_mtime:
```
mschaef-da: I guess I get what you're trying to achieve with this, but two control flow paths in a test setup method make me nervous about test stability. Pretty sure this can be simplified to always do the same thing. (And if we find that it's a performance issue, can be corrected.)

python/tests/unit/test_package_loading.py
```diff
@@ -0,0 +1,72 @@
...
+        # Attempt to load only the DALF into the Sandbox; this will cause the PackageService to
+        # return a nonsensical result, the lack of an exception being thrown signals that we can
+        # tolerate this condition
+        d = {}
+
+        with sandbox(DALF_FILE) as proc:
+            network = Network()
+            network.set_config(url=proc.url)
+            client = network.aio_party('TestParty')
+            client.add_ledger_ready(lambda event: setitem(d, 'metadata', event.package_store))
```
mschaef-da: This is a nice API that DAZL presents.

python/dazl/protocols/v1/pb_parse_metadata.py
```diff
@@ -82,11 +89,27 @@ def find_dependencies(
...
+
     m_pb = {}
     sorted_package_ids = toposort_flatten(dependencies)
     for package_id in sorted_package_ids:
-        m_pb[package_id] = metadatas_pb[package_id]
-    return m_pb
+        if package_id not in unresolvable_package_ids:
+            required_package = metadatas_pb.get(package_id)
+            if required_package is not None:
+                m_pb[package_id] = required_package
```
mschaef-da: Given that you have both a then clause and an else clause, I'd be tempted to do `is None` rather than `is not None`.

python/dazl/protocols/v1/pb_parse_metadata.py
```diff
@@ -82,11 +89,27 @@ def find_dependencies(
...
+            else:
+                LOG.warning('Failed to find a package %r', package_id)
+
+    return ArchiveDependencyResult(
+        sorted_archives=m_pb,
+        unresolvable_archives={pkg_id: metadatas_pb[pkg_id] for pkg_id in unresolvable_package_ids})
```
mschaef-da: This seems like the sort of thing that should be logged (or maybe reported as an error?).

python/tests/unit/test_package_loading.py
da-tanabe: Yeah fair enough—I tried with and without and the perf was nearly identical. Removed.

python/tests/unit/test_package_loading.py
da-tanabe: heh…yep, added a helper method `find_nearest_ancestor` (which I want for other reasons anyway).

python/dazl/protocols/v1/pb_parse_metadata.py
da-tanabe: Added some log statements.

PR #18, opened 2019-06-27
python: Package upload support

da-tanabe:
Add support for uploading packages through the Package Management Service API.

Most of the changes in this PR are actually around configuration management, specifically making `--parties` optional as it is not a required parameter when uploading packages. Also fully removes provisional support for uploading through the Cloudbox REST API.


python/dazl/client/_network_client_impl.py
```diff
@@ -312,6 +310,18 @@ def emit_event(self, data: BaseEvent):
 
     # endregion
 
+    async def connect_anonymous(self):
```
mschaef-da: This seems like a bigger deal to me than just for package uploads.

What happens when you have an anonymous connection and attempt to perform a ledger create/exercise?

da-tanabe: You wouldn't be able to—the underlying `_PartyClientImpl` has those methods instead.

PR #20, opened 2019-07-08
python: Make connection failures retryable.

da-tanabe:


python/dazl/protocols/v1/grpc.py
```diff
@@ -168,23 +168,40 @@ def grpc_upload_package(connection: 'GRPCv1Connection', dar_contents: bytes) ->
...
-            LOG.info("No gRPC Ledger Identity Service found!")
-            return None
-        LOG.exception('An unexpected error occurred when trying to fetch the ledger identity.')
-        raise
 
-    return response.ledger_id
+    start_time = datetime.utcnow()
+    connect_timeout = stub.context.options.connect_timeout
+
+    while connect_timeout is None or (datetime.utcnow() - start_time) < connect_timeout:
```
mschaef-da: How is is this throttled if l191 doesn’t take time?

python/dazl/protocols/autodetect.py
```diff
@@ -164,7 +166,13 @@ def _main(self) -> None:
                         self._context.run_on_loop(lambda: self._ledger_future.set_result(metadata))
                     elif metadata is not None:
                         LOG.warning('The network monitor thread emitted multiple metadata!')
+        except (UserTerminateRequest, ConnectionTimeoutError) as ex:
+            # re-raise these, but they are "known" errors so a stack trace in the logs would just
```
mschaef-da: Will the user see at least the exception type here?

python/dazl/protocols/v1/grpc.py
```diff
@@ -168,23 +168,40 @@ def grpc_upload_package(connection: 'GRPCv1Connection', dar_contents: bytes) ->
...
+        if stub.context.run_state.terminate_requested:
+            raise UserTerminateRequest()
+        if stub.closed:
+            raise Exception('connection closed')
+
+        try:
+            response = stub.ledger_identity_service.GetLedgerIdentity(G.GetLedgerIdentityRequest())
+        except RpcError as ex:
+            details_str = ex.details()
+            if details_str in ('DNS resolution failed', 'failed to connect to all addresses'):
```
mschaef-da: Ok, but yuck.

da-tanabe: Yeah, sadly there isn't a more comprehensive gRPC API. Everything from these kinds of connection failures to a missing gRPC service comes back with a status code of `UNAVAILABLE`.

python/dazl/protocols/v1/grpc.py
da-tanabe: There's a `sleep` on line 196. Also because the gRPC API is a blocking API (at least in Python), the sleep below is a thread sleep. :-/

python/dazl/cli/metadata.py
```diff
@@ -57,10 +58,16 @@ def execute_static_metadata(files: Collection[str], options: PrettyOptions) -> i
...
-        network = Network()
-        network.set_config(config)
-        network.run_until_complete(_main(network, options))
-        return 0
+        try:
+            network = Network()
+            network.set_config(config)
+            network.run_until_complete(_main(network, options))
+            return 0
+        except UserTerminateRequest:
```
da-tanabe: @mschaef-da example of where the non-logged errors are handled by "user" code.

python/dazl/protocols/autodetect.py
da-tanabe: I made it the responsibility of the caller in these two cases. So in the case of the first one (CTRL+C), no, because I didn't want CTRL+C to emit a crazy looking (and completely irrelevant) stack trace; commented above in the PR where logging would happen if the calling code decided to log.

python/dazl/protocols/v1/grpc.py
mschaef-da: That's 1msec, right? I'd rather it be longer, but okay.

da-tanabe: Nope, 1 second.

PR #22, opened 2019-07-12
python: Make using the Active Contract Set service _not_ the default

da-tanabe:
Make using the Active Contract Set service _not_ the default (appears to be buggy).

Will need to come up with better repro steps and file a bug on the `daml` repo, but the Active Contract Set service is returning far too many "extra" contracts. Using the ACS is now configurable, and I have flipped it off by default.


python/dazl/client/config.py
```diff
@@ -132,6 +132,11 @@ class _NetworkConfig:
         param_type=SECONDS_TYPE,
         default_value=1)
 
+    use_acs_service: bool = config_field(
```
mschaef-da: Not happy with the introduction of another config flag, but okay, given the ACS service issues.

da-tanabe: Yeah you and me both. :-/

PR #25, opened 2019-07-16
python: Make package upload blocking more "synchronous"

da-tanabe:
Wait until DAR uploads are done _and_ packages are reported locally before returning.

### DIFF
--- a/python/dazl/client/_network_client_impl.py
+++ b/python/dazl/client/_network_client_impl.py
@@ -24,6 +24,7 @@ from ..model.reading import InitEvent, ReadyEvent, BaseEvent, EventKey
 from ..protocols import LedgerNetwork
 from ..protocols.autodetect import AutodetectLedgerNetwork
 from ..util.asyncio_util import execute_in_loop, completed, Invoker, safe_create_future
+from ..util.dar import get_dar_package_ids
 from ..util.events import CallbackManager
 from ..util.prim_types import to_timedelta, TimeDeltaConvertible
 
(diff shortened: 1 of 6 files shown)


python/dazl/client/_network_client_impl.py
```diff
@@ -269,9 +270,10 @@ def simple_metadata(self, timeout: Optional[float] = 30.0) -> LedgerMetadata:
         pool = await self._pool_init
         return await pool.ledger()
 
-    async def ensure_package(self, contents: bytes, timeout: TimeDeltaConvertible) -> None:
+    async def upload_package(self, contents: bytes, timeout: 'TimeDeltaConvertible') -> None:
         """
-        Ensure packages specified by the given byte array are loaded on the remote server.
+        Ensure packages specified by the given byte array are loaded on the remote server. This
+        method only returns once packages are reported by the package services.
```
mschaef-da: Is there any utility in retaining the non-blocking variant too? I'm thinking no, but maybe in larger ledger scenarios where ensuring the presence of a DAR takes a lot of time. (Multi-domain, maybe?)

python/dazl/util/io.py
```diff
@@ -7,17 +7,53 @@
...
+def get_bytes(src):
     """
-    Read the contents of a file as ``bytes``. If ``file_path`` is ``None``, then ``None`` is
-    returned instead.
+    Read the contents of a file as ``bytes``. If ``src`` is ``None``, then ``None`` is returned
+    instead.
     """
+    if src is None:
+        return None
+    elif isinstance(src, str):
```
mschaef-da: Doesn't Python provide some sort of better way to do type dispatch? (ie, something similar to this: https://clojure.org/reference/multimethods )

python/dazl/client/_network_client_impl.py
da-tanabe: I'm going to leave it out for now, if only because if you want "non-blocking" behavior, then you can simply fall back to more advanced use of `async`/`await` to more precisely control the async flow.

python/dazl/util/io.py
da-tanabe: It does: https://docs.python.org/3/library/functools.html

I haven't actually used it extensively because it _only_ dispatches on type and it's also more lines of code.

PR #26, opened 2019-07-18
python: Add support for raising an event when a packages are added.

da-tanabe:
Introduce `@ledger_packages_added`, which allows a dazl app to be notified when packages changed (rather than needing to poll on the "outside").

Many of the changes in this PR are actually around cleaning up some of the boilerplate around type signatures with respect to event handling.


python/dazl/client/_events.py
```diff
@@ -0,0 +1,36 @@
...
+
+EventHandler = Callable[[E], EventHandlerResponse]
+AEventHandler = Callable[[E], Union[EventHandlerResponse, Awaitable[EventHandlerResponse]]]
+
+
+EventHandlerDecorator = Callable[[EventHandler[E]], EventHandler[E]]
+AEventHandlerDecorator = Callable[[AEventHandler[E]], AEventHandler[E]]
+
+
+def fluentize(fn: 'Callable[[T], None]', *args, **kwargs) -> 'Callable[[T], T]':
```
mschaef-da: nice.

python/dazl/model/reading.py
```diff
@@ -157,6 +157,14 @@ class ContractArchiveEvent(ContractEvent):
     """
 
 
+@dataclass(frozen=True)
+class PackagesAddedEvent(BaseEvent):
+    """
```
mschaef-da: We need to stream this up to the UI to detect new templates, etc.

PR #27, opened 2019-07-19
python: Fix DAML model deprecation warnings.

da-tanabe:
Just long-overdue renaming DAML 1.1 `Integer` to DAML 1.2 `Int`.


python/_template/Main.daml
```diff
@@ -37,10 +37,10 @@ template AuthorRole
               address
               content
 
-      nonconsuming CreateIntegerLetter : ContractId UnsortedLetter
+      nonconsuming CreateIntLetter : ContractId UnsortedLetter
```
mschaef-da: I guess the theory behind this change is that these aren't really integers?

da-tanabe: Most likely…at one point DAML `Integers` did not have limits, but that's of course not the case any more.

PR #33, opened 2019-08-01
python: Introduce "Bot" as a first-class concept

da-tanabe:
Introduce "Bot" as a first-class concept that can be individually addressed.

Minor changes to API (mostly around methods that formerly returned nothing, actually returning something), and major changes to how event handlers are scheduled under the hood.


python/dazl/client/_network_client_impl.py
```diff
@@ -182,6 +182,8 @@ def background_main():
             except:
                 LOG.exception('The main event loop died!')
 
+            LOG.info('The main event loop has finished.1')
```
mschaef-da: Why is there a numeral 1 at the end of this?

python/dazl/client/_network_client_impl.py
```diff
@@ -253,7 +255,7 @@ def party_impl(self, party: Party, ctor: 'Callable[[_PartyClientImpl], T]') -> T
                 mm[party] = inst
             return inst
 
-    def parties(self) -> Collection[Party]:
+    def parties(self) -> 'Collection[Party]':
```
mschaef-da: What significance is there to the quotes?

python/dazl/client/bots.py
```diff
@@ -0,0 +1,315 @@
+# Copyright (c) 2019 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
```
mschaef-da: I really need to put copyrights in the Ui code...

python/dazl/client/_party_client_impl.py
```diff
@@ -115,7 +114,9 @@ def set_time(self, new_datetime) -> Awaitable[None]:
     # region Event Handler Management
 
     # noinspection PyShadowingBuiltins
-    def add_event_handler(self, key: str, handler, filter, context):
+    def add_event_handler(
+            self, key: str, handler: BotCallback, filter: 'Optional[BotFilter]', context: Any) \
```
mschaef-da: Python’s relationship with whitespace truly is an unnatural thing...

python/dazl/client/bots.py
```diff
@@ -0,0 +1,315 @@
...
+                # the main queue contains either events we have not yet processed yet or ``None``
+                # markers that merely indicate running status should be "re-checked"
+                invocation = await self._queue.get()
+                if invocation is not None:
+                    try:
+                        fut = ensure_future(self._handle_event(invocation.event))
+                        propagate(fut, invocation.future)
+                        await invocation.future
+                    except Exception:
+                        LOG.exception('An event handler in a bot has thrown an exception!')
```
mschaef-da: Maybe log at least a few details of the instigating event.

python/dazl/client/bots.py
```diff
@@ -0,0 +1,315 @@
...
+        queued up while the bot was paused. When this queue is fully drained, the state is changed
+        to ``RUNNING``.
+        """
+        LOG.info('Resuming the bot thread for party %r...', self.party)
+        self._state = BotState.RESUMING
+        self._queue.put_nowait(None)
+
+    def stop(self):
+        """
+        Permanently stop this bot. If you need to be able to "restart" a stopped bot, use
```
mschaef-da: Why would I ever want to pause a bot?

python/dazl/client/bots.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def notify(self, event: 'BaseEvent') -> 'Awaitable[None]':
+        """
+        Notifies handler(s) associated with this bot that the given event has occurred. Note that
+        this notification is asynchronous: in other words, event handlers will not have processed
+        this event by the time this function returns.
+
+        :param event: The event to raise.
+        """
+        if not isinstance(event, BaseEvent):
```
mschaef-da: There has to be a standard library function for this idiom...

python/dazl/client/_network_client_impl.py
da-tanabe: Oops. :-)

python/dazl/client/_network_client_impl.py
da-tanabe: Nothing other than consistency…I'm slowly moving the whole codebase of all the things to this model.

It's technically only needed for forward references (https://www.python.org/dev/peps/pep-0484/#forward-references) but I don't want to have reformat code just because I may have moved things around, so I just use quotes for consistency now.

python/dazl/client/_party_client_impl.py
da-tanabe: hahahaha…yeah kind of a mess.

python/dazl/client/bots.py
da-tanabe: yep good call…done.

python/dazl/client/bots.py
da-tanabe: This is in support of start/stop functionality in the project:DABL UI.

python/dazl/client/bots.py
da-tanabe: The standard Python idiom would be to let it fail (not even _do_ a check), but because this is async code and the actual failure would happen far from the call site, I do this check proactively.

python/dazl/client/bots.py
mschaef-da: I am somewhat sorry I asked. 😊

PR #36, opened 2019-08-06
python: Expose a management endpoint for configuring bots at runtime.

da-tanabe:
* Expose a management endpoint for configuring bots at runtime.


python/dazl/model/core.py
```diff
@@ -178,6 +185,19 @@ class DazlWarning(Warning):
     """
 
 
+class DazlPartyMissingError(DazlError):
```
mschaef-da: If we're going to define an error class that says 'this party is missing', we should add a field to the class that lets a catch identify which party is missing. (Without resorting to parsing the message string.)

python/dazl/model/core.py
```diff
@@ -178,6 +185,19 @@ class DazlWarning(Warning):
...
+    """
+    Error raised when a party or some information about a party is requested, and that party is not
+    found.
+    """
+
+
+class DazlImportError(ImportError, DazlError):
```
mschaef-da: Same sort of thing here... if we're going to have the class, there needs to be a way to capture which dependency could not be found.

python/dazl/model/core.py
```diff
@@ -178,6 +185,19 @@ class DazlWarning(Warning):
...
+    """
+    Import error raised when an optional dependency could not be found.
```
mschaef-da: Is it right that it's an *error* to be unable to find an *optional* dependency?

python/dazl/server/management.py
```diff
@@ -0,0 +1,96 @@
...
+Endpoints for managing parties/bots connected via a dazl client.
+"""
+
+from dataclasses import asdict, dataclass
+from typing import Collection, Sequence, TYPE_CHECKING
+
+from ..client import Bot, _NetworkImpl
+from ..model.core import DazlImportError, Party, SourceLocation
+
+if TYPE_CHECKING:
```
mschaef-da: That's cool.

python/dazl/client/_network_client_impl.py
```diff
@@ -266,6 +318,19 @@ def party_impls(self) -> 'Collection[_PartyClientImpl]':
         with self._lock:
             return list(self._party_impls.values())
 
+    def find_bot(self, bot_id) -> 'Optional[Bot]':
+        """
+        Return the bot of the specified ID, or ``None`` if no matching bot is found.
+        """
+        # TODO: Improve on the performance of this by indexing on bot ID. Linear iteration isn't
+        #  likely to cause a performance problem for a few dozen bots, but for apps with thousands
+        #  of event handlers, this could start to be a problem.
```
mschaef-da: I'd remove the todo until we know that the issue will actually happen. Optimization should be driven by demand and profiling rather than speculative comments, and if this ever becomes a problem, the profiler won't let us miss it.

python/dazl/model/core.py
mschaef-da: Looking further down in the code, this looks like it's thrown when there's an attempt to spin up a server and the HTTP server dependency is not present within the Python environment. If that's the scenario, then this seems more like an 'unsupported feature' or 'bad configuration' exception where the reason just happens to be a missing dependency.

python/dazl/server/management.py
mschaef-da: I think.

python/dazl/server/management.py
```diff
@@ -0,0 +1,96 @@
...
+    entries: 'Sequence[BotInfoEntry]'
+
+
+@dataclass(frozen=True)
+class BotInfoEntry:
+    event_key: str
+    source_location: 'SourceLocation'
+
+
+def build_routes(network_impl: '_NetworkImpl') -> 'Collection[web.RouteDef]':
```
mschaef-da: What role does DAZL play in our architecture? Until this PR, I"d viewed it exclusively as a python client library for the ledger. (Which doesn't really jibe with the fact that it can host an HTTP server.)

da-tanabe: We'd need something _else_ on top of dazl (like some kind of app framework), and in practice, people would need to use that instead of just using dazl directly.

Although it's a good question—a dazl app either has to be able to expose some mechanism for controlling it from the outside, or a well-known endpoint that a dazl app plugs into to pump information into it. I still prefer the former, because I think some basic things like health checks and metrics work better when they originate from the application rather than the application having to publish.

da-tanabe: I guess I'm viewing this HTTP port as really more a "debugging port" than a full-blown HTTP server with all of those implications.

python/dazl/server/management.py
da-tanabe: It gets around circular import errors, so that's nice…I guess.

Would be more awesome if circular import errors weren't a thing I had to contend with.

python/dazl/model/core.py
da-tanabe: It's fairly typical for `ImportError` to be raised by code that requires an optional dependency but doesn't have one specified (an example from SQLAlchemy: https://github.com/sqlalchemy/sqlalchemy/blob/6f270fb0e344016ce526e0a1ecb3e3de9ffd1e3b/lib/sqlalchemy/dialects/sqlite/pysqlcipher.py#L100); the only reason I'm not just "letting it go" (which would be the typical Python approach) is that I'm trying to get better about making sure that only `DazlError` subclasses are raised from dazl code, and that all other classes of errors and exceptions are things that users should explicitly handle.

python/dazl/client/_network_client_impl.py
```diff
@@ -241,12 +244,61 @@ def global_impl(self, ctor: 'Callable[[_NetworkImpl], T]') -> T:
                 self._global_impls[ctor] = inst
             return inst
 
-    def party_impl(self, party: Party, ctor: 'Callable[[_PartyClientImpl], T]') -> T:
+    @overload
+    def party_impl(
+            self, party: Party, ctor: None = None,
+            if_missing: Literal[CREATE_IF_MISSING, EXCEPTION_IF_MISSING] = CREATE_IF_MISSING) -> \
```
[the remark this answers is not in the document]
da-tanabe: I thought long and hard about this one, but eventually decided to go with a "When in Rome" approach. The few others that use `Literal` types in Python seem to be doing it this way; it doesn't help that the syntax for using literal types is _significantly_ clunkier than it is in TypeScript (TypeScript just lets you say `'something'` where a type is expected, where as Python requires these awful `Literal['something']` wrappers).

PR #39, opened 2019-08-14
python: Fix errors in processing TextMap

da-tanabe:
* Fix errors in processing TextMap.
* Support for SDK 0.13.18.
* Dropped support for pre-0.13.0 types.
* Pre-compile all DAML files used in the scope of tests (constantly rebuilding DAML files was taking an unacceptably long time).


python/dazl/damlsdk/package.py
```diff
@@ -27,7 +27,10 @@ def package(options: 'PackageOptions', component: 'Optional[str]' = None) -> 'Pr
         raise ValueError(f'Unknown packaging component: {component}')
 
     LOG.info('Using %s to create a DAR package...', name)
-    args = [path / 'da-hs-damlc-app', *_damlc_package_options(options)]
+    jars = list(path.glob('*.jar'))
```
mschaef-da: What does this have to do with `TextMap`?

python/dazl/protocols/v1/pb_parse_event.py
```diff
@@ -443,40 +445,29 @@ def to_variant(context: TypeEvaluationContext, tt: Type, variant: 'G.Variant') -
...
-            return to_natural_type(child_context.append_path('?'), ot.type_parameter, variant.value)
-
-    return type_evaluate_dispatch_default_error(
-        on_optional=process_optional,
-        on_map=process_map,
-        on_variant=process_variant)(context, tt)
+    return type_evaluate_dispatch_default_error(on_variant=process)(context, tt)
+
+
+def to_map(context: TypeEvaluationContext, tt: Type, map_pb: 'G.Map') -> Any:
```
mschaef-da: Can we use a more specific return type? A function `to_map` that returns `Any` seems a bit disjoint to me.

python/dazl/damlsdk/package.py
da-tanabe: The tests have been stuck on SDK 0.13.10 for a while because the mechanism used to fetch damlc had changed. This change allows me to go up to higher versions of the SDK.

python/dazl/protocols/v1/pb_parse_event.py
da-tanabe: Yep, fair…changed.

PR #41, opened 2019-08-14
python: Clean up some type signatures.

da-tanabe:
Change some type signatures to make some of the type APIs easier to code against.


python/dazl/model/types.py
```diff
@@ -73,22 +75,19 @@ def names(self):
...
+        on_type_app: Callable[['TypeApp'], T_co],
+        on_scalar: Callable[['ScalarType'], T_co],
+        on_contract_id: Callable[['ContractIdType'], T_co],
+        on_optional: Callable[['OptionalType'], T_co],
+        on_list: Callable[['ListType'], T_co],
+        on_text_map: 'Callable[[TextMapType], T_co]',
+        on_record: Callable[['RecordType'], T_co],
+        on_variant: Callable[['VariantType'], T_co],
+        on_enum: 'Callable[[EnumType], T_co]',
+        on_unsupported: Callable[['UnsupportedType'], T_co]) -> Callable[['Type'], T_co]:
```
chiara-da: why is it being called T_co?

da-tanabe: Seems to be the standard way the Python world conveys covariance: https://docs.python.org/3/library/typing.html#typing.Iterable

Issue #45, opened 2019-08-28
Add documentation to github pages


lucianojoublanc-da:
The docs are generated with Sphinx. There are couple of howto's on the web explaining how to make the doc available through github pages. TL;DR put generated html under `docs/`.

Ideally this should be autogenerated with CI - this may be trickier.

Would be nice to reuse the theme from https://docs.daml.com for consistency.

lucianojoublanc-da:
I've got a simple solution using gh-pages (see PR #46 ), however the styling doesn't appear to work - either that or it's just very ugly :rofl: . I don't know if it's the best use of my time trying to resolve this. Somebody with knowledge of CSS and how Sphinx works may be better placed to tackle this - but I'm happy to give it a go if that is the consensus.

PR #46, opened 2019-08-28
Add github-pages & update landing page docs

lucianojoublanc-da:
This adds a root `docs/` folder where documentation is now generated, for #45 . Generating this automatically with CI is more complicated and I couldn't get this working after an hour of tinkering. The downside of this is that we need to commit changes to the `docs/` folder manually for the time being, after building docs locally (using `make build`, or `cd python && make docs). Would be nice to switch to CI doc builds in the future.

To enable the github pages docs, **you'll need to enable them in github settings**, toward the bottom of the page. The docs will then be hosted online at https://digital-asset.github.io/dazl-client

Also note that there appears to be something wrong with the CSS styles. Would appreciate if somebody with more knowledge of CSS/Sphinx could check that out.

I updated the README/index files to reflect that the build system is now [Poetry](https://poetry.eustace.io/), and also added a note that the library can be installed using `pip`, for the casual user.


lucianojoublanc-da:
You can preview the docs at https://lucianojoublanc-da.github.io/dazl-client

da-tanabe:
Closing this fork-based PR in favor of a branch-based one (#103).

PR #55, opened 2019-10-27
python: Make dazl a little less susceptible to crashes

da-tanabe:
A few odds and ends:

* Make dazl a little less susceptible to crashes of its main thread.
* Make `typing-extensions` a dependency only installed for Python 3.7 and older.
* Fix a deprecation warning in importing `collections`.


python/dazl/protocols/_base.py
```diff
@@ -140,7 +140,7 @@ def __init__(self,
         self.executor = executor if executor is not None else ThreadPoolExecutor(max_workers=25)
 
     def run_in_background(self, func, *args) -> Future:
-        return ensure_future(self.loop.run_in_executor(self.executor, func, *args))
+        return shield(ensure_future(self.loop.run_in_executor(self.executor, func, *args)))
```
mschaef-da: This is really nice...

da-tanabe:
`exc_info=True` makes the logger log the full exception (just as it would have before), but without the crash.

PR #56, opened 2019-10-30
python: Support for DAML-LF.dev (SDK 0.13.32).

da-tanabe:
* Revise Protobuf downloading to keep up with recent changes to how DAML-LF proto files are published.
* Add support for DAML-LF 1.dev (SDK 0.13.32); in particular, this means adding support for interned strings.


python/dazl/damlast/pb_parse.py
```diff
@@ -546,7 +595,9 @@ def parse_DefDataType_Fields(self, pb) -> 'DefDataType.Fields':
             fields=tuple(self.parse_FieldWithType(field) for field in pb.fields))
 
     def parse_DefDataType_EnumConstructors(self, pb) -> 'DefDataType.EnumConstructors':
-        return DefDataType.EnumConstructors(constructors=tuple(pb.constructors))
+        ctors_1 = tuple(pb.constructors)
```
mschaef-da: image

python/dazl/damlast/daml_lf_1.py
```diff
@@ -1307,6 +1405,14 @@ class BuiltinFunction(Enum):
     DIV_DECIMAL = 3
     ROUND_DECIMAL = 6
 
+    ADD_NUMERIC = 107
```
mschaef-da: Do we have a disassembler for these trees? Something that'll take a choice and dump it out in a readable format?

da-tanabe: Not really…at least not in `dazl` (although there is a fair amount of code trying to nudge in that direction).

da-tanabe:
Agreed…and the TypeScript codebase around DAML types is similarly bloated. I've been thinking about trying to codegen my way out of this mess for a while, but that in itself is also a heavy lift. :-/

PR #66, opened 2019-12-20
python: Completely switch over to using Pytest-style tests.

da-tanabe:
Switch to using pytest-style tests (we are already using `pytest` as the test runner).

The primary impetus for this change is a major change to the way that the Sandbox is launched from within dazl tests due to changes in the way that Sandbox has started shipping since DAML SDK 0.13.39:
* dazl will now launch `daml sandbox` instead of trying to launch the Sandbox .jar from a well-known path. The `dazl.sandbox()` utility method is not compatible DAML SDK 0.13.39 and will be deprecated soon.
* The tests in dazl will finally run against official SDK images instead of selectively trying to download sandbox.jar from Bintray and launching it from there.
* *dazl will use a Pytest fixture in tests to talk to a long-running Sandbox, instead of controlling Sandbox processes*. The CircleCI image that runs Python tests will subsequently no longer need `java`, as this can now be considered an implementation detail of the DAML SDK Docker image.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -15,7 +15,7 @@ Rich Python bindings for accessing Ledger API-based applications.
 Requirements
 ------------
 * Python 3.6+
-* [Pipenv](https://pipenv.readthedocs.io/en/latest/)
+* [Poetry](https://python-poetry.org/)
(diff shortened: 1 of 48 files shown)


mschaef-da:
Looks good (and mostly like a mechanical translation).

PR #72, opened 2020-03-19
python: Fix a bug that caused ready() calls to perpetually hang under certain circumstances

da-tanabe:
Fix a bug that caused ready() calls to perpetually hang if a connection was initially opened, but dropped before the ACS could fully be read.

* Factor out scheduling logic to its own package.
* Remove excess wrapper functions and classes from scheduling concerns.

This is _mostly_ a mechanical change to remove some extra layers in the code where they weren't needed; the most salient change is moving `Future` creation to a function that tracks outstanding `Future`'s, and is responsible for cancelling them all when the `Invoker` is shut down.


python/dazl/scheduler/_invoker.py
```diff
@@ -0,0 +1,131 @@
...
+import signal
+
+from ._base import RunLevel
+from ..util.asyncio_util import safe_create_future, execute_in_loop
+from ..util.prim_types import to_timedelta
+
+DEFAULT_TIMEOUT = timedelta(seconds=30)
+
+
+class Invoker:
```
mschaef-da: This seems like it might be much more generally useful than just here. (ie: The integrations stuff can use it.) Between that and enough other gaps that we've seen in the Python ecosystem, I'm wondering if there might be room for some sort of DA or DABL specific Python utility library.

python/dazl/scheduler/_base.py
```diff
@@ -0,0 +1,30 @@
...
+
+from ..util.enum import OrderedEnum
+
+
+class RunLevel(OrderedEnum):
+    """
+    The possible running modes for an :class:`Invoker`. Invokers generally progress from
+    lower-numbered run levels to higher-numbered run levels.
+    """
+    RUN_FOREVER = 0
```
mschaef-da: Is this really a run _level_ or a run _state_ in a state machine? The comment gives me the impression that this is a state, but the name makes It seem more like the Unix concept of system runlevel (which has always seemed more like a just a way to identify a specific configuration of the world.)

PR #78, opened 2020-04-12
python: Allow for offsets to be non-parseable (DAML SDK 1.0 and later).

da-tanabe:
As part of the upcoming DAML SDK 1.0 release, the string format for offsets have changed; they are now more straightforward to compare (they don't need to be parsed at all to be sorted), but the format _has_ changed somewhat.

Allow for parse errors in the format of transaction offsets to simply fall back to comparing the natural sort order of those strings, as this is all that needs to be done for DAML SDK 1.0 and later.


python/dazl/model/reading.py
```diff
@@ -317,11 +317,16 @@ def max_offset(offsets: 'Collection[str]') -> 'Optional[str]':
...
-            return int(components[1])
+        try:
+            components = value.split('-', 3)
+            if len(components) == 1:
+                return int(value)
+            elif len(components) >= 1:
+                return int(components[1])
+        except ValueError:
+            # newer versions of the DAML SDK no longer return offsets in a parseable format, but
+            # the strings themselves are naturally comparable
```
mschaef-da, line 330: Are there useful semantics to this comparison? (ie: Sorting a block hash isn't terribly interesting. in terms of offset height.)

da-tanabe: Yup—the new format guarantees that these opaque strings are actually comparable (whereas before parsing was required to "find" the sortable part that determined relative ordering of two transactions).

PR #83, opened 2020-04-21
python: Ignore archive events that correspond to unobserved create events

da-tanabe:
Ignore archive events that correspond to create events that have never been observed in dazl's active contract set store.


python/dazl/client/state.py
```diff
@@ -129,10 +129,12 @@ def handle_create(self, event: ContractCreateEvent) -> None:
         self._queries = [q for q in self._queries if not q.check_ready(self)]
 
     def handle_archive(self, event: ContractArchiveEvent) -> None:
+        existing_data = self._data.get(event.cid.contract_id)
+        effective_at = existing_data.effective_at if existing_data is not None else None
         self._data[event.cid.contract_id] = ContractContextualData(
             cid=event.cid,
             cdata=None,
-            effective_at=self._data[event.cid.contract_id].effective_at,
```
da-tanabe, line 135: @albertleung-da @lima-da @maxdeliso-da this is where we were seeing `KeyError`s before when trying to dump the state of our ledger.

PR #93, opened 2020-07-29
python: Support formatting decimals that are very small.

da-tanabe:
The default `Decimal.__str__` implementation renders very small Decimal numbers in scientific notation, which does not conform to the gRPC Ledger API spec with respect to decimal serialization:

https://github.com/digital-asset/daml/blob/39b6e47a7e5dda6d5aad71cfca47d8d2df1722d6/ledger-api/grpc-definitions/com/daml/ledger/api/v1/value.proto#L42-L47

This change ensures that `Decimal` values are properly serialized.


python/dazl/protocols/v1/pb_ser_command.py
```diff
@@ -48,6 +49,11 @@ def as_identifier(tref: 'TypeReference') -> 'G.Identifier':
     return identifier
 
 
+def as_ledegr_api_decimal(d: 'Decimal') -> str:
```
albertleung-da, line 52: typo on ledger

da-tanabe: oops thx

PR #106, opened 2020-08-06
python: Update the dependencies list.

da-tanabe:
Removed `pyyaml` as a dependency (`dazl` no longer uses it in any way), and move many of the remaining dependencies to optional dependencies that can be selectively opted in if those features are used.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -10,38 +10,40 @@ homepage = "https://github.com/digital-asset/dazl-client"
 keywords = ["daml", "blockchain", "dlt", "distributed ledger", "digital asset"]
 
 [tool.poetry.dependencies]
-python = ">=3.6"
-aiohttp = "*"
+python = "^3.6"
+aiohttp = { version = "*", optional = true }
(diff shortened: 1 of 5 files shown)


python/pyproject.toml
```diff
@@ -10,38 +10,40 @@ homepage = "https://github.com/digital-asset/dazl-client"
...
-grpcio = ">=1.20.1"
-oauthlib = "*"
+google-auth = { version = "*", optional = true }
+grpcio = ">=1.29.1"
+oauthlib = { version = "*", optional = true }
 prometheus_client = { version = "*", optional = true }
-protobuf = ">=3.8.0"
+protobuf = ">=3.12.0"
 pygments = { version = "*", optional = true }
-pyyaml = "*"
```
mschaef-da, line 22: Yay!

da-tanabe: Thanks for raising! The last of the code that used YAML was removed earlier this week, so this _definitely_ doesn't belong here any more! :-)

PR #108, opened 2020-08-06
python: Next-gen dazl connection API

da-tanabe:
## dazl v8 API

Introduces a new API that embraces multi-party subscriptions, and modernizes and simplifies the implementation.

```py
async def main(token):
    async with dazl.connect(url, token=token) as conn:
        await conn.create("Some:Thing", { ... })
        async with conn.stream("Some:Thing") as stream:
            async for event in stream:
                print(event)
                if i_feel_like_it:
                    break
                elif event.cdata["name"] == "Batman":
                    await buy_batmobile(event)
```

## Fixes

`dazl` turns four years old in May 2021, and many of the design decisions that it was originally built with (even predating the Daml gRPC Ledger API as introduced in October 2018) no longer apply:

### No longer consolidates to a single transaction stream per `Party`

* `SubmitAndWait`: older versions of the ledger API _only_ had a `CommandSubmissionService.Submit` call, which required clients to listen to `TransactionStream` events in order to know whether or not a command succeeded or not. Establishing independent transaction streams _per command submission_ is obviously an unscalable approach, so `dazl` instead listens to the transaction stream and doles out completion events internally.
* `ActiveContractSetService` didn't originally exist, so `dazl` has a tradition of reading from the transaction stream, even when it doesn't necessarily need to. This was partially remediated in #10/#16, but `dazl` is still very much predicated on reading from the transaction stream, and incorporation of the `ActiveContractSetService` is somewhat incomplete. By virtue of the single-stream-per-`Party` design, the ACS is also similarly one-per-`Party`.
* The introduction of contract keys removes the need for applications to maintain a lot of in-state memory. In practice, modern Daml applications don't need to hold on to the ACS in memory any more, but `dazl` makes applications generally pay for maintaining an ACS nonetheless.
* Exercise result values have been sent over the Ledger API for a while, but since command submission in `dazl` is _only_ asynchronous, correlating exercise results back to call sites was very challenging.
* Most importantly, multi-party submissions makes it impossible for `dazl`'s singular transaction stream-per-`Party` model from being workable.

### gRPC and `asyncio`

* The gRPC library for Python has recently incorporated `asyncio` support (https://github.com/grpc/proposal/pull/155)!  This removes the need for the complicated internals in `dazl` devoted to trying to keep Python thread count under control. Every transaction stream subscription required at least _three_ Python threads to maintain, and this was another reason that `dazl` tried to "conserve" transaction subscriptions.

### HTTP JSON API support/TypeScript library parity

* The HTTP JSON API exposes a subset of the gRPC Ledger API that supports almost all of what `dazl` currently exposes a library, and it _should_ be possible for an application written against `dazl` to talk to either API.
* Authenticated ledger support is not well-documented, and the current design does not support token refreshing at all.
* The TypeScript bindings expose a very similar API to `dazl`'s; with some minor symbol renames, the APIs are effectively identical.

##

This does _not_ yet include the compatibility layer to help ease the transition from the old API to the new one.

Some new low-level utility libraries have been added: see #172, #175, #176, #179

Due to the size of the change, some long-deprecated APIs were dropped as a prerequisite to this work: #155, #159

### DIFF
--- a/python/dazl/ledger/grpc/conn_aio.py
+++ b/python/dazl/ledger/grpc/conn_aio.py
@@ -3,6 +3,8 @@
 """
 This module contains the mapping between gRPC calls and Python/dazl types.
 """
+from __future__ import annotations
+
 import asyncio
 from typing import AbstractSet, Any, AsyncIterable, Collection, Mapping, Optional, Sequence, Union
 import uuid
@@ -52,7 +54,7 @@ from ..._gen.com.daml.ledger.api.v1.transaction_service_pb2 import (
 )
 from ..._gen.com.daml.ledger.api.v1.transaction_service_pb2_grpc import TransactionServiceStub
 from ...damlast.daml_lf_1 import PackageRef, TypeConName
-from ...prim import ContractData, ContractId, Party
+from ...prim import LEDGER_STRING_REGEX, ContractData, ContractId, Party
 from ...query import Query
 from ..api_types import ArchiveEvent, Boundary, Command, CreateEvent, ExerciseResponse, PartyInfo
 from ..config import Config
@@ -66,27 +68,28 @@ __all__ = ["Connection"]
 
 
 class Connection:
-    def __init__(self, config: "Config"):
+    def __init__(self, config: Config):
         self._config = config
+        self._logger = config.logger
         self._channel = create_channel(config)
         self._codec = Codec(self)
 
     @property
-    def config(self) -> "Config":
+    def config(self) -> Config:
         return self._config
 
     @property
-    def channel(self) -> "Channel":
+    def channel(self) -> Channel:
         """
         Provides access to the underlying gRPC channel.
         """
         return self._channel
 
     @property
-    def codec(self) -> "Codec":
+    def codec(self) -> Codec:
         return self._codec
 
-    async def __aenter__(self) -> "Connection":
+    async def __aenter__(self) -> Connection:
         await self.open()
         return self
 
@@ -104,9 +107,10 @@ class Connection:
             stub = LedgerIdentityServiceStub(self._channel)
             response = await stub.GetLedgerIdentity(G_GetLedgerIdentityRequest())
             if isinstance(self._config.access, PropertyBasedAccessConfig):
+                self._logger.info("Connected to gRPC Ledger API, ledger ID: %s", response.ledger_id)
                 self._config.access.ledger_id = response.ledger_id
             else:
-                raise ValueError("token-based access must supply tokens that provide ledger ID")
+                raise ValueError("when using token-based access, the token must contain ledger ID")
 
     async def close(self) -> None:
         """
@@ -119,10 +123,10 @@ class Connection:
 
     async def do_commands(
         self,
-        commands: "Union[Command, Sequence[Command]]",
+        commands: Union[Command, Sequence[Command]],
         *,
-        workflow_id: "Optional[str]" = None,
-        command_id: "Optional[str]" = None,
+        workflow_id: Optional[str] = None,
+        command_id: Optional[str] = None,
     ) -> "None":
         """
         Submit one or more commands to the Ledger API.
@@ -161,12 +165,12 @@ class Connection:
 
     async def create(
         self,
-        template_id: "Union[str, TypeConName]",
-        payload: "ContractData",
+        template_id: Union[str, TypeConName],
+        payload: ContractData,
         *,
-        workflow_id: "Optional[str]" = None,
-        command_id: "Optional[str]" = None,
-    ) -> "CreateEvent":
+        workflow_id: Optional[str] = None,
+        command_id: Optional[str] = None,
+    ) -> CreateEvent:
         """
         Create a contract for a given template.
 
@@ -195,13 +199,13 @@ class Connection:
 
     async def exercise(
         self,
-        contract_id: "ContractId",
+        contract_id: ContractId,
         choice_name: str,
-        argument: "Optional[ContractData]" = None,
+        argument: Optional[ContractData] = None,
         *,
-        workflow_id: "Optional[str]" = None,
-        command_id: "Optional[str]" = None,
-    ) -> "ExerciseResponse":
+        workflow_id: Optional[str] = None,
+        command_id: Optional[str] = None,
+    ) -> ExerciseResponse:
         """
         Exercise a choice on a contract identified by its contract ID.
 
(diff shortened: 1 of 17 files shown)


python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -0,0 +1,484 @@
...
+        :return:
+            A stream of responses, where each response contains one or more events at a particular
+            offset.
+        """
+        filters = await self.conn.codec.encode_filters(self._queries)
+        filters_by_party = {party: filters for party in self.conn.config.access.read_as}
+        tx_filter_pb = G_TransactionFilter(filters_by_party=filters_by_party)
+
+        try:
+            offset = None
```
mschaef-da, line 417: It may be worth a comment that `_acs_events` always emits a boundary with an offset.

python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -0,0 +1,484 @@
...
+                if isinstance(event, CreateEvent):
+                    await self._emit_create(event)
+                elif isinstance(event, Boundary):
+                    offset = event.offset
+                    await self._emit_boundary(event)
+                else:
+                    warnings.warn(f"Received an unknown event: {event}", ProtocolWarning)
+                yield event
+
+            if self._continue_stream:
```
mschaef-da, line 428: A couple things:

* If `offset` is `None` here, then something is wrong, no? Does Python have an assert thing that can enforce this invariant?
* If the ledger API doesn't guarantee to always be able to produce archived contracts, how long is `offset` guaranteed to be a valid input for `_tx_events`? I'm assuming long enough that we don't have to worry here, but not sure.

python/docs/migrating.rst
```diff
@@ -49,8 +49,184 @@ Commands such as ``dazl ls``, ``dazl tail``, or options provided for your applic
...
+* Daml return values from exercise choices
+
+  Exercise return values have been available from the Ledger API since Daml was open sourced back
+  in April 2019. dazl's API has finally been updated to take this feature into account.
+
+* "Unsubscribing" from streams
+
+  ``dazl`` has never had a straightforward way of abandoning event callbacks that were no longer
+  needed. The new API makes stream lifecycle more explicit and the responsibility of the user of the
+  library. Disposing of streams is now simpler to reason about.
```
mschaef-da, line 91: Nice.

mschaef-da (commented):
A couple minor comments, but generally this looks very good and I am looking forward to using it.

Given the scope of this change, my suggestion (which may have actually come from you yourself)  is to release the as a beta v8, let us get some time with it over the next couple of weeks, and then cut the official v8 once we're sure it works the way we want.

python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -0,0 +1,484 @@
...
+        await self.close()
+
+    async def open(self) -> None:
+        """
+        Does final validation of the token, including possibly fetching the ledger ID if it is not
+        yet known.
+        """
+        if not self._config.access.ledger_id:
+            # most calls require a ledger ID; if it wasn't supplied as part of our token or we were
+            # never given a token in the first place, fetch the ledger ID from the destination
```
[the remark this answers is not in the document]
da-tanabe: Ya good point—will add some `INFO` logs here.

python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -0,0 +1,484 @@
...
+    @property
+    def config(self) -> "Config":
+        return self._config
+
+    @property
+    def channel(self) -> "Channel":
+        """
+        Provides access to the underlying gRPC channel.
+        """
+        return self._channel
```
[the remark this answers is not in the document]
da-tanabe: Sent an internal-only code link to you offline…but I've needed direct access to the connection for other reasons in the codebase. The new API actually removes the need to use the connection directly in the aforementioned linked URL, but it may still have long-term value to keep exposed.

python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -0,0 +1,484 @@
...
+                    await self._emit_boundary(event)
+                else:
+                    warnings.warn(f"Received an unknown event: {event}", ProtocolWarning)
+                yield event
+
+            if self._continue_stream:
+                # now start returning events as they come off the transaction stream; note this
+                # stream will never naturally close, so it's on the caller to call close() or to
+                # otherwise exit our current context
+                async for event in self._tx_events(tx_filter_pb, offset):
```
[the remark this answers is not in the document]
da-tanabe: Yup exactly! This is one of the things I'm most excited about with respect to the new API— _every_ interaction with the ledger provides an obvious place to wrap with `try`/`except`, whereas the more callback-heavy approach made this much more difficult/impossible.

python/dazl/ledger/grpc/conn_aio.py
da-tanabe: Good point—shall add.

python/dazl/ledger/grpc/conn_aio.py
da-tanabe: On the first point, not necessarily—`offset is None` means that you're connecting to an empty ledger and are interested in hearing create/archive events going forward.

On the second point, I'm not sure…but there isn't much we can do about it from a client-side perspective.

python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -0,0 +1,484 @@
...
+        """
+        if not self._config.access.ledger_id:
+            # most calls require a ledger ID; if it wasn't supplied as part of our token or we were
+            # never given a token in the first place, fetch the ledger ID from the destination
+            stub = LedgerIdentityServiceStub(self._channel)
+            response = await stub.GetLedgerIdentity(G_GetLedgerIdentityRequest())
+            if isinstance(self._config.access, PropertyBasedAccessConfig):
+                self._config.access.ledger_id = response.ledger_id
+            else:
+                raise ValueError("token-based access must supply tokens that provide ledger ID")
```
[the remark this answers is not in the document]
da-tanabe: I rephrased this to perhaps to make it clearer, but the story is nuanced:
* For property-based access on the gRPC Ledger API, the ledger ID can be queried. This is how `dazl` works today, and why you don't really ever need to specify ledger IDs
* For property-based access on the HTTP JSON API, `dazl` mints its own unsigned tokens, because even in an un-authed configuration, clients still identify themselves to the server by using tokens. _Here_, a ledger ID _cannot_ be inferred, so it must be specified.
* For token-based access on _either_ gRPC Ledger API or HTTP JSON API, the ledger ID is "part of the spec", so it's required for the token to be considered valid. (And because the token is generally signed, the client can't modify specific claims in the token, particularly ledger ID).

python/dazl/ledger/grpc/conn_aio.py
mschaef-da: Yeah... this is a big improvement.

mschaef-da (approved):

PR #110, opened 2020-08-18
python: Reinstate support for some deprecated template formats.

da-tanabe:
This restores the ability to resolve template names of the form `PKG:ENTITY.MOD` (removed in #104), as this has broken a number of `dazl` packages attempting to upgrade from 6.x.

The other deprecated and removed features do not appear to have been as widely used.


python/dazl/model/lookup.py
```diff
@@ -68,12 +68,28 @@ def template_reverse_globs(primary_only: bool, package_id: str, type_name: str)
...
+                    yield f'{package_id}:{m}:{e}'
+                    if primary_only:
+                        return
             yield f'{package_id}:{type_name}'
             if primary_only:
                 return
         if not primary_only or type_name == '*':
             yield f'{package_id}:*'
     if type_name != '*':
+        if ':' not in type_name:
```
mschaef-da, line 85: Is it easily possible to log a deprecation warning here? That's about all I'd suggest changing.

da-tanabe: Unfortunately this area of the code will get called many times—I'd need to add the deprecation warning closer to the call site (where event handlers are registered). Adding it here would only spam the logs. :-/

PR #122, opened 2020-09-14
go: Add low-level Ledger API and DAML-LF bindings.

da-tanabe:
go: Add code-generated Protobuf/gRPC files for the Ledger API and DAML-LF.


jordan-da (approved):
so it begins

mschaef-da (approved):

PR #125, opened 2020-09-22
python: Improve startup performance of dazl-based applications

da-tanabe:
* Improve performance of reading only package IDs from a DAR. In particular do _not_ fully parse a DAR's constituent DALFs when all that is requested is the set of package IDs (which, ironically, is used to optimize whether or not to fetch and parse DARs from the `PackageService`).

* Make dazl properly respect the `use_acs_service` setting. This configuration should have been applied per-`Party`, but inadvertently `dazl` used a `use_acs_service` value of `False` regardless of what it was set to.
  This _may_ be a breaking change for applications that rely on `record_time` for contracts, but this is a relatively uncommon use case.


mschaef-da (approved):
Looks like a nice improvement.

PR #137, opened 2020-10-04
python: Improvements to logging infrastructure.

da-tanabe:
python: Improvements to logging infrastructure.
* Move the logging code to `dazl/_logging.py`.
* Deprecate `dazl.setup_default_logger`; applications should instead prefer to configure their own loggers (simply calling `logging.basicConfig()` is a good-enough alternative for most applications).
* Use `pytest`'s native support for logging the output of loggers, rather than the slightly more indirect route of writing log output to stdout, and then having `pytest` look at stdout.
* Before #66, `dazl` used Python's built-in `unittest` framework, which has no integration with Python's built-in `logging` framework. `dazl.util.io.StdoutStreamWrapper` existed solely to support `logger`/`unittest` integration, so it has been dropped.
* Add `LOG.verbose_timed`, `LOG.debug_timed`, and `LOG.info_timed`, all which produce context managers that can be used to easily log potentially long-running blocks of code.

### DIFF
--- a/_fixtures/src/post-office/integration-test.py
+++ b/_fixtures/src/post-office/integration-test.py
@@ -9,9 +9,9 @@ from collections import defaultdict
 from datetime import datetime, timedelta
 from pathlib import Path
 
-from dazl import create, exercise, setup_default_logger, Network, write_acs
+from dazl import create, exercise, Network, write_acs
 
-setup_default_logger()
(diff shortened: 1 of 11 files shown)


mschaef-da (approved):
NIce.... I like this one. :-)

PR #138, opened 2020-10-05
python: Consolidate primitive type handling

da-tanabe:
python: Consolidate the code that handles conversion to/from primitive Python types to a new package, `dazl.prim`.
* Deprecate `dazl.model.core.ContractId` and replace it with `dazl.prim.ContractId`.
* Deprecate `dazl.util.prim_types` and move much of it to `dazl.prim`.
* Deprecate `dazl.protocols.v0.json_ser_command` and move much of it to `dazl.prim.json`.


mschaef-da (approved):
Is this a breaking change?

da-tanabe:
It was very much a goal to _not_ make this a breaking change; all public symbols are still present, and emit `DeprecationWarning` before calling the function from its new place. Additionally as I was developing, I moved the symbols first and left the rest of the code calling the deprecated functions; the tests still passed (albeit with thousands of warnings :-P ) so I'm reasonably confident that if anyone were calling these (arguably) internal functions directly, they'll get the appropriate warning, but their code will continue to work.

In particular, `dazl.prim.ContractId` is a supertype of `dazl.model.core.ContractId`, which means that anyone constructing `dazl.model.core.ContractId` can still use those instances; furthermore, dazl will still create instances of `dazl.model.core.ContractId` until the type is removed.

PR #141, opened 2020-10-05
python: Introduce dazl.damlast.pkgfile.DarFile

da-tanabe:
Deprecate `dazl.util.dar.DarFile` and its API for fetching data as a `PackageStore`.

The new `DarFile` is a bit lighter-weight and does not do any internal caching, and parsing happens _synchronously_. Asynchronous parsing (to avoid blocking the main thread) will be handled by a new dedicated package loader in an upcoming PR.

### DIFF
--- a/_fixtures/src/post-office/integration-test.py
+++ b/_fixtures/src/post-office/integration-test.py
@@ -11,8 +11,6 @@ from pathlib import Path
 
 from dazl import create, exercise, Network, write_acs
 
-logging.basicConfig()
-
 LOG = logging.getLogger('integration_test')
 
 POSTMAN_PARTY = 'Postman'
@@ -138,6 +136,9 @@ def _main():
     """
     import argparse
     import subprocess
+
+    logging.basicConfig(level=logging.INFO)
+
     from dazl.util import ProcessLogger, find_free_port, kill_process_tree, wait_for_process_port
 
     argparser = argparse.ArgumentParser()
--- a/python/dazl/damlast/errors.py
+++ b/python/dazl/damlast/errors.py
@@ -1,0 +1,47 @@
+# Copyright (c) 2017-2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Error types
+-----------
+
+.. autoclass:: PackageNotFoundError
+.. autoclass:: NameNotFoundError
+"""
+
+from typing import Any, TYPE_CHECKING
+
+from ..model.core import DazlError
+
+if TYPE_CHECKING:
+    from . import daml_lf_1
+
+__all__ = ['PackageNotFoundError', 'NameNotFoundError']
+
+
+class PackageNotFoundError(DazlError):
+    """
+    Raised when a :class:`Package` was expected but could not be found.
+
+    This error can also be raised in places where types/values are being resolved; it is an
+    indication that the operation MAY succeed if the package of the specified ID is fetched and
+    the original option retried.
+    """
+
+    def __init__(self, ref: 'daml_lf_1.PackageRef'):
+        super().__init__(ref)
+        self.ref = ref
+
+
+class NameNotFoundError(DazlError):
+    """
+    Raised when a :class:`daml_lf_1.DefDataType`, :class:`daml_lf_1.DefValue`, or
+    :class:`daml_lf_1.DefTemplate` of a specific name was expected, but could not be found.
+
+    Typically this error is raised when the *package ID* is valid, but the name is not. Because
+    package IDs are immutable, this error is not normally retryable.
+    """
+
+    def __init__(self, ref: 'Any'):
+        super().__init__(ref)
+        self.ref = ref
--- a/python/dazl/damlast/pkgfile.py
+++ b/python/dazl/damlast/pkgfile.py
@@ -1,0 +1,183 @@
+# Copyright (c) 2017-2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from io import BytesIO
+from os import PathLike
+from pathlib import Path
+from typing import AbstractSet, BinaryIO, Collection, Generator, Mapping, Optional, Union
+from zipfile import ZipFile
+
+from .daml_lf_1 import Archive, Package, PackageRef
+from .parse import parse_archive
+
+from .._gen.com.daml.daml_lf_dev import daml_lf_pb2 as pb
+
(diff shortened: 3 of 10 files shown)


_fixtures/src/post-office/integration-test.py
```diff
@@ -138,6 +136,9 @@ def _main():
     """
     import argparse
     import subprocess
+
+    logging.basicConfig(level=logging.INFO)
```
mschaef-da, line 140: Much better.

python/dazl/damlast/errors.py
```diff
@@ -0,0 +1,47 @@
...
+
+from ..model.core import DazlError
+
+if TYPE_CHECKING:
+    from . import daml_lf_1
+
+__all__ = ['PackageNotFoundError', 'NameNotFoundError']
+
+
+class PackageNotFoundError(DazlError):
```
mschaef-da, line 22: I've never been completely convinced about the utility of separate exception classes for different errors, although I don't suppose it hurts all that much.

python/dazl/damlast/pkgfile.py
```diff
@@ -0,0 +1,183 @@
...
+
+        elif isinstance(dar, PathLike):
+            self.filename = dar.__fspath__()
+            self._buf = None
+            self._zip = ZipFile(self.filename)
+
+        elif isinstance(dar, bytes):
+            self.filename = None
+            self._buf = BytesIO(dar)
+            self._zip = ZipFile(self._buf)
```
mschaef-da, line 50: Nice.

python/dazl/damlast/pkgfile.py
```diff
@@ -0,0 +1,183 @@
...
+
+        elif hasattr(dar, 'read'):
+            self.filename = None
+            self._buf = None
+            self._zip = ZipFile(dar)
```
mschaef-da, line 55: What case does this cover that the other don't?

python/dazl/damlast/pkgfile.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        names = self._zip.namelist()
+        if 'META-INF/MANIFEST.MF' in names:
+            manifest_bytes = self._zip.read('META-INF/MANIFEST.MF')
+            manifest = {}
+            for line in manifest_bytes.decode('utf-8').splitlines():
+                print(line)
+                name, _, value = line.partition(':')
+                manifest[name] = value.strip()
+            return manifest
```
mschaef-da, line 83: This defines the type of a manifest as `Mapping[str, str]` and then uses some generic string manipulation to parse it out. I don't know that this is the PR for the work, but a more formal manifest parser and data type might be warranted.

python/dazl/damlast/pkgfile.py
```diff
@@ -0,0 +1,183 @@
...
+    def manifest(self) -> 'Optional[Mapping[str, str]]':
+        """
+        Return the contents of the manifest of this DAR.
+        """
+        names = self._zip.namelist()
+        if 'META-INF/MANIFEST.MF' in names:
+            manifest_bytes = self._zip.read('META-INF/MANIFEST.MF')
+            manifest = {}
+            for line in manifest_bytes.decode('utf-8').splitlines():
+                print(line)
```
mschaef-da, line 80: Is the print intentional?

python/dazl/damlast/pkgfile.py
```diff
@@ -0,0 +1,183 @@
...
+                name, _, value = line.partition(':')
+                manifest[name] = value.strip()
+            return manifest
+        else:
+            return None
+
+    def sdk_version(self) -> 'Optional[str]':
+        """
+        Return the SDK version used to compile this dar (if this information is available).
+        """
```
mschaef-da, line 90: If version is really a string, it might be useful to describe the expected format of that string, so consumers of this API can do things other than print, equality check, etc.

python/dazl/damlast/pkgfile.py
```diff
@@ -0,0 +1,183 @@
...
+        Return the :class:`Package` corresponding to the specified :class:`PackageRef`.
+
+        This function is somewhat inefficient as it must check every DALF for the specified package
+        ID. The :method:`DarFile.archives` function should be used if it is known that all archives
+        in a DAR are to be accessed, or use a :class:`CachedDarFile` instance if random access of
+        packages by package ID are needed.
+        """
+        for a in self._pb_archives():
+            if a.hash == package_id:
```
mschaef-da, line 113: Does this need to defend against the case there there are multiple archives with the same `package_id`? (Even if that is a signal of an erroneously formed archive.)

mschaef-da (approved):

python/dazl/damlast/errors.py
dtanabe: Ah…I actually will be using this soon! Stay tuned. :-)

python/dazl/damlast/pkgfile.py
dtanabe: Eek! Good catch—will remove.

python/dazl/damlast/pkgfile.py
dtanabe: File-like objects…believe it or not, Python has no formal interface for the object type returned by `open()`! (https://docs.python.org/3/glossary.html#term-file-object)

dtanabe: (also added an explanatory comment)

python/dazl/damlast/pkgfile.py
da-tanabe: Unfortunately it's not actually that well-defined (https://github.com/digital-asset/daml/issues/329); the ideal way to document this would be a link to something on docs.daml.com.

python/dazl/damlast/pkgfile.py
da-tanabe: Perhaps…it may be something to revisit if/when `META-INF/MANIFEST.MF` carries more valuable information. My personal opinion is that the information in the manifest file should eventually find their ways to `.dalf` files; indeed, some data now exists in DAML-LF today:

https://github.com/digital-asset/daml/blob/eb68e680f649bc7c2b6b2c3b8a4c7a664f52ecec/daml-lf/archive/src/main/protobuf/com/digitalasset/daml_lf_1_8/daml_lf_1.proto#L1423-L1426

So in that light I'm hopeful that one day this kind of parsing can simply be dropped.

python/dazl/damlast/pkgfile.py
da-tanabe: That's a tough question…a few months/years ago I'd say yes, but as a general effort to try to slim down the size of the library, I'll just cop out and say that sending dazl invalid DARs is undefined behavior. :-P

PR #142, opened 2020-10-06
python: Introduce SymbolLookup and PackageLoader

da-tanabe:
Introduce `SymbolLookup` and `PackageLoader`, the intended replacements for `PackageStore`.

`PackageLookup` and `SymbolLookup` aren't wired into any of the infrastructure of `dazl` yet; that will be coming in a subsequent PR.

This change actually manages to be backwards/forwards compatible because `PackageLookup`, a lazier version of `PackageStore`, doesn't actually contain any packages until they are requested; this means if for whatever reason a version of `dazl` were cut on this commit, the `PackageStore` would continue to try to remain up-to-date in its currently "almost working" state, and the `PackageLookup` would simply remain empty.

### DIFF
--- a/python/dazl/client/pkg_loader.py
+++ b/python/dazl/client/pkg_loader.py
@@ -1,0 +1,209 @@
+# Copyright (c) 2017-2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import sys
+from asyncio import ensure_future, get_event_loop, gather, wait_for, sleep
+from datetime import timedelta
+from concurrent.futures import ThreadPoolExecutor
+from typing import AbstractSet, Awaitable, Callable, Dict, Set, TypeVar
+
+from .. import LOG
+from ..damlast.daml_lf_1 import Archive, Package, PackageRef
+from ..damlast.pkgfile import Dar
+from ..damlast.errors import PackageNotFoundError, NameNotFoundError
+from ..damlast.lookup import MultiPackageLookup
+from ..model.core import DazlError
+from ..model.lookup import validate_template
+
+if sys.version_info >= (3, 7):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
+
+
+__all__ = ['SyncPackageService', 'PackageLoader']
+
+
+T = TypeVar('T')
+
+DEFAULT_TIMEOUT = timedelta(seconds=30)
+
+
+class SyncPackageService(Protocol):
+    """
+    A service that synchronously provides package information.
+    """
+
+    def package_bytes(self, package_id: 'PackageRef') -> bytes:
+        raise NotImplementedError('SyncPackageService.package_bytes requires an implementation')
+
+    def package_ids(self) -> 'AbstractSet[PackageRef]':
+        raise NotImplementedError('SyncPackageService.package_ids requires an implementation')
+
+
+class PackageLoader:
+    """
+    Loader for packages from a remote PackageService.
+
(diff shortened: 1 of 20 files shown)


python/dazl/client/pkg_loader.py
```diff
@@ -0,0 +1,201 @@
...
+        :param fn: A function to invoke.
+        :return: The result of that function.
+        """
+        failed_types = set()  # type: Set[str]
+        failed_packages = set()  # type: Set[PackageRef]
+        while True:
+            try:
+                return fn()
+
+            except PackageNotFoundError as ex:
```
mschaef-da, line 86: Ah.... now I see.

mschaef-da: I tend to think these various lookup failure cases need at least debug level logging, so they can be traced through.

python/dazl/client/pkg_loader.py
```diff
@@ -0,0 +1,201 @@
...
+        self._conn = conn
+        self._timeout = timeout
+        self._loading_futs = dict()  # type: Dict[PackageRef, Awaitable[Package]]
+        self._parsing_futs = dict()  # type: Dict[PackageRef, Awaitable[Archive]]
+        self._executor = ThreadPoolExecutor(3)
+
+    def set_connection(self, conn):
+        self._conn = conn
+
+    async def do_with_retry(self, fn: 'Callable[[], T]') -> 'T':
```
mschaef-da, line 68: This seems like a very specific kind of retry... specific enough that a more specific name is warranted.

python/dazl/client/pkg_loader.py
```diff
@@ -0,0 +1,201 @@
...
+
+                pkg_id, _ = validate_template(ex.ref)
+                if pkg_id == '*':
+                    # we don't know what package contains this type, so we have no
+                    # choice but to look in all known packages
+                    failed_types.add(ex.ref)
+                    await self.load_all()
+                else:
+                    # we know what package this type comes from, but it did not contain
+                    # the required type
```
mschaef-da, line 111: Yeah... all these comments are useful to describe what the branches mean, but to me that implies the ability to access control flow information through this using debug log message would be useful.

mschaef-da: (Yes.... I'm using one PR comment to agree with another of my comments. You only need to worry when I start disagreeing with myself.)

python/dazl/client/pkg_loader.py
```diff
@@ -0,0 +1,201 @@
...
+
+        loop = get_event_loop()
+        conn = self._conn
+        if conn is None:
+            raise DazlError('a connection is not configured')
+
+        archive_bytes = await wait_for(
+            self.__fetch_package_bytes(conn, package_id), timeout=self._timeout.total_seconds())
+
+        LOG.info("Bytes loaded for package: %s", package_id)
```
mschaef-da, line 162: Might be interesting to see a count of bytes here.

mschaef-da (approved):

python/dazl/client/pkg_loader.py
da-tanabe: How's about VERBOSE level then? It's worth pointing out that in the new code flow, exceptions here are actually more of a usual flow than before.

(Incidentally I very much struggled with that design decision as it is very atypical for me to write code that _relies_ on exceptions, but the alternative would have been to make the new parsing/serialization code fully async, which I think would have both made the code harder to follow as well as incur a potential performance penalty.)

da-tanabe: Actually good call on this front—this prompted me to add another unit test to make sure that failed lookups are truly terminal.

PR #143, opened 2020-10-08
python: Fix a package load race condition

da-tanabe:
Fix a race condition that could cause failures when trying to use packages around the same time they are uploaded to the ledger.

This is the one that switches the primary encode/decoder to the new `ValueMapper` introduced in #144, and uses the `SymbolLookup`/`PackageLoader` combination from #142.

A few long-standing TODOs in the guts of dazl have also now been dropped, particularly around ordering of when packages get loaded, and inopportune polling as it relates to trying to make sure the appropriate packages are available at the appropriate time (which turns out to be impossible to do).

### DIFF
--- a/python/dazl/client/state.py
+++ b/python/dazl/client/state.py
@@ -1,26 +1,25 @@
 # Copyright (c) 2017-2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-import asyncio
 import warnings
-from asyncio import Future
+from asyncio import Future, sleep
 from collections import defaultdict
 from typing import Awaitable, Dict, List, Optional, Union, Collection, cast
 
 from ..client._reader_match import is_match
 from ..damlast.daml_lf_1 import TypeConName
+from ..damlast.protocols import SymbolLookup
 from ..model.core import ContractId, ContractsState, ContractMatch, \
     UnknownTemplateWarning, ContractContextualData, ContractContextualDataCollection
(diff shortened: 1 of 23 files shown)


python/dazl/client/state.py
```diff
@@ -54,60 +53,50 @@ def read_full(self,
...
-        matching_templates = store.resolve_template_type(template_name)
-        if matching_templates:
-            return {tt.con: self._tcdata[tt.con] for tt in matching_templates}
-        else:
+    async def read_async(
+            self, template_name: str, match: ContractMatch = None, min_count: int = 1) \
+            -> 'ContractsState':
+        # Fetch matching names; we may need to wait for a package to be fetched in order to do this.
+        # Note that this loop may never terminate; in this case, this behavior is actually
+        # acceptable because it is what enables us to wait indefinitely for the creation of a
```
mschaef-da, line 67: How sure are you that we don't need a timeout here? Indefinate loop has me a bit concerned (particularly when viewing this from the point of view of an external non-DA user that hasn't spent any time thinking through how DAZL works).

mschaef-da: (From that external point of view, I'd almost prefer it to blow up with an exception and a reference to the timeout that I can adjust if I know that I can be a bit more patient.)

mschaef-da (approved):

python/dazl/client/state.py
da-tanabe: Added some logging to illustrate to callers if/when this behavior happens to make it more obvious if this is happening.

PR #144, opened 2020-10-08
python: Introduce ValueMapper

da-tanabe:
Introduce ValueMapper, a replacement for Serializer on the write side and a replacement for pb_parse_event (which was gRPC only) on the read side.

Unlike `Serializer`, `ValueMapper` uses DAML-LF `dazl.damlast.daml_lf_1.Type` objects instead of `dazl`'s original `dazl.model.types.Type` object, which means extending dazl to handle new types introduced in DAML-LF in the future will be much easier. Additionally the new mappers have no implicit dependency on `PackageStore` fully-resolved types behavior as the mappers are built on `SymbolLookup`, which has well-defined behavior for how it handles a missing package lookup.

This PR was split out of #143 simply because #143 was too large. This PR's focus is on introducing the new implementation into the codebase (along with its tests); #143 will be the actual cutover to using this code.

### DIFF
--- a/python/dazl/values/canonical.py
+++ b/python/dazl/values/canonical.py
@@ -1,0 +1,146 @@
+# Copyright (c) 2017-2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Mapping, Tuple
+
+from .context import Context
+from .mapper import ValueMapper
+from ..damlast.daml_lf_1 import DefDataType, Type
+from ..prim import to_bool, to_date, to_record, to_variant, to_int, to_party, to_datetime, to_str, \
+    to_decimal
+
+
+class CanonicalMapper(ValueMapper):
+    """
+    A mapper that canonicalizes values. If values are already in canonical form, this is essentially
+    the identity mapper.
+
+    For container primitives (Optional, List, Map, etc), records, and variants, values are recursed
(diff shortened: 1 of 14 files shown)


python/dazl/values/canonical.py
```diff
@@ -0,0 +1,146 @@
...
+
+        expected_keys = frozenset(fld.field for fld in record.fields)
+        actual_keys = frozenset(orig_mapping)
+        if actual_keys.issuperset(expected_keys):
+            if actual_keys != expected_keys:
+                # Earlier versions of dazl were more tolerant of extra fields. To keep backwards
+                # compatibility, we'll emit a warning, though this may become an exception
+                # eventually.
+                context.value_warn(
+                    obj, f'extra fields: {", ".join(sorted(actual_keys - expected_keys))}')
```
mschaef-da, line 40: Nice touch.

python/dazl/values/context.py
```diff
@@ -0,0 +1,311 @@
...
+                elif item_type.prim.prim == PrimType.GENMAP:
+                    return self.mapper.prim_gen_map(
+                        self, item_type.prim.args[0], item_type.prim.args[1], obj)
+
+            raise ValueError(f'could not process type: {item_type}')
+
+        except ValueError as ex:
+            # re-raise ValueErrors with a more descriptive string that includes the current context
+            # path
+            raise ValueError(f'value at {".".join(self.path)} has an invalid value: {obj}') from ex
```
mschaef-da, line 113: Nice.

mschaef-da (approved):

PR #154, opened 2021-02-13
python: Enforce formatting as part of the build.

da-tanabe:
python:
* Enforce isort/black formatting as part of the build.
* Make isort's sorting behavior compatible with black's settings, so that `isort --check-only` does not fail.
* Format gRPC/Protobuf-generated files; we could configure the build to ignore them, but there isn't really a compelling reason to do so. (I didn't originally do this as part of #153 thinking that exclusion of these directories was going to be more straightforward to configure than it turned out to be.)
* Remove some PyCharm comments around excluding `grpc`/`protobuf` imports. (I originally thought these were `grpcio`/`protobuf` library bugs, but apparently it was actually PyCharm, and fixed some time within the last few months.)


mschaef-da (approved):
Looks really good. I am a particular fan of isort...  very nice. (Back in the day, I used to do this: https://github.com/mschaef/.emacs.d/blob/master/lisp/javap.el#L481 )

PR #156, opened 2021-02-26
python: Command hierarchy revamp

da-tanabe:
python: Prepare the `Command` classes for the new v8 API (#108).

* Rename `CreateCommand`/`CreateAndExerciseCommand`/`ExerciseCommand`/`ExerciseByKeyCommand` fields to match the properties as they exist in the JavaScript ledger bindings (https://docs.daml.com/app-dev/bindings-ts/daml-ledger/classes/_index_.ledger.html).
* Split these classes into `dazl.protocols.commands.*Command`, and deprecated counterparts in `dazl.clients.commands.*Command` that are meant to help with the v8 transition but will be removed in v9.
* Remove usage of `dataclass` in `Command`, as dataclasses don't have a particularly straightforward way of deprecating fields.

### DIFF
--- a/python/dazl/client/_party_client_impl.py
+++ b/python/dazl/client/_party_client_impl.py
@@ -1,6 +1,5 @@
 # Copyright (c) 2017-2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-
 from asyncio import Future, ensure_future, gather, get_event_loop, wait
 from concurrent.futures import ALL_COMPLETED
 from dataclasses import asdict, dataclass, field, fields, replace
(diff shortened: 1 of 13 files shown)


python/dazl/client/_party_client_impl.py
```diff
@@ -481,7 +476,9 @@ def write_commands(
         """
         if workflow_id is None:
             workflow_id = uuid.uuid4().hex
-        cb = CommandBuilder.coerce(commands, atomic_default=True)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            cb = CommandBuilder.coerce(commands, atomic_default=True)
```
mschaef-da, line 481: I've said this before, but I think I really like this pattern for managing warnings.

mschaef-da (approved):

python/dazl/client/_party_client_impl.py
da-tanabe: Agreed—this is something that Python does rather well.

PR #159, opened 2021-03-04
python: Drop dazl's deprecated type hierarchy

da-tanabe:
python:
* Drop dazl's deprecated type hierarchy in `dazl.model.types` and `dazl.model.types_store`.
* Move some of `dazl.models.events` into `dazl.client.events`, as that code is only used there.

This is the first of a few steps to remove `dazl.models` altogether; _all_ of the types in this package will either move to `dazl.protocols` (as it pertains to the v8 API) or `dazl.client` (as it pertains to the v5 API). When the transition is complete and there are _no_ remaining references to `dazl.models`, a new compatibility layer will be reintroduced in `dazl.models` that merely imports `dazl.client` symbols so that existing imports do not break. This can only easily be done once `dazl.models` is ultimately removed, as right now `dazl.models` is imported in too many places and introducing the compatibility layer too early would result in circular dependencies.


python/dazl/client/bots.py
```diff
@@ -46,6 +46,13 @@
 BotFilter = Callable[[E], bool]
 
 
+@dataclass(frozen=True)
+class SourceLocation:
```
mschaef-da, line 50: What is this used for again? This seems like another reasonable candidate for potential future deprecation.

python/dazl/client/events.py
```diff
@@ -0,0 +1,193 @@
...
+            return on_offset(event)
+        elif isinstance(event, PackagesAddedEvent):
+            return on_packages_added(event)
+        else:
+            raise ValueError(f"unknown subclass of BaseEvent: {event!r}")
+
+    return handle
+
+
+def template_reverse_globs(primary_only: bool, package_id: str, type_name: str) -> "Iterator[str]":
```
mschaef-da, line 61: Are there tests for this? This is tricky enough and close enough to the user API that it might be worthwhile to add them if not.

python/dazl/damlast/types.py
```diff
@@ -1,21 +1,11 @@
...
-import warnings
+from typing import Callable
 
-from ..model.types import Type as OldType
-from ..util.typing import safe_cast
 from ._base import T
 from .daml_lf_1 import PrimType, Type
 
 
-def var(var_: str) -> "Type":
```
mschaef-da, line 12: Good to see deprecated code removed.

mschaef-da (approved):

python/dazl/client/bots.py
da-tanabe: Just for bots—and yeah, definitely on the list to be deprecated.

python/dazl/client/events.py
da-tanabe: Yup this is definitely covered by tests. In dazl 7 I had _intentionally_ changed it: I reverted back in 7.0.1, but didn't give people enough time to switch over to the current syntax…but the tests asserted what I had intended 🤦 .

Issue #109, opened 2020-08-09
network.simple_client appears to ignore or drop config settings 


nycnewman:
If I set config to have parameters for ssl and oauth, these parameters appears to be dropped when SimplePartyClient implementation is created via network.coimple_client('Alice')

Need better documentation on correct approach to configuring (mutual) TLS and oAuth settings to get appropriate JWT token from an oAuth provider.

PR #178, opened 2021-03-16
python: Add the new v8 config objects, with docs.

da-tanabe:
Add classes for the new config system that will be used with the v8 API (#109).

Some goals of this implementation:
* Support multiparty submissions more effectively, particularly add support for specifying `--act-as` and `--read-as`
* Be less "magical" than the current implementation (see [dazl/util/config.py](https://github.com/digital-asset/dazl-client/blob/master/python/dazl/util/config.py) and [dazl/util/config_meta.py](https://github.com/digital-asset/dazl-client/blob/master/python/dazl/util/config_meta.py))
* `Network.set_config(**kwargs)` takes a lot of options, but because it's defined as "kwargs", the documentation doesn't show up in an IDE which also hurts discoverability.
* Be easier to plug into client applications; `dazl.run` has been deprecated for a while, but because it handles configuring `argparse` for you, people have had trouble getting off of it
* Produce less spew when running something like `dazl ls --help`; there are a ton of options and they are not well-organized and there is a fair amount of visual noise.

New `dazl ls --help` (as produced by a test)
```
usage: sample app [url config] [access config] [ssl/tls config]

  # connect to a local sandbox acting as Alice and Bob, with Carol read-only
  # (localhost is the default if no url/host is specified)
  sample app -u Alice,Bob -r Carol

  # connect to myledger.daml.app over SSL/TLS using the access token in "token.txt"
  sample app --url https://myledger.daml.app/ --oauth-token-file token.txt

Basic options:
  --url URL                                URL of the gRPC Ledger API/HTTP JSON API (cannot be used with
                                           --host/--port)
  --host/-H HOST                           ledger host (cannot be used with --url; default: localhost)
  --port/-p PORT                           ledger port (cannot be used with --url; default: based on
                                           scheme/host)
  --scheme {http,https,grpc,grpcs}         scheme/protocol (cannot be used with --url)

Access configuration (property-based):
  --act-as/-u PARTY [PARTY ...]            act-as parties
  --read-as/-r PARTY [PARTY ...]           read-as parties
  --admin                                  Add admin flag (HTTP JSON API only)
  --ledger-id LEDGER_ID, -l LEDGER_ID      ledger ID (required on HTTP JSON API)
  --application-name NAME                  application name

Access configuration (token-based):
  --oauth-token TOKEN | --oauth-token-file FILE 
                                           OAuth2 bearer token

Advanced connection options:
  --connect-timeout SECONDS                number of seconds before giving up on a connection (default:
                                           30.0)
  --plaintext                              force the connection to use plaintext instead of SSL/TLS
  --cert CERT | --cert-file FILE           client certificate (mutual TLS)
  --cert-key CERT | --cert-key-file FILE   client certificate private key (mutual TLS)
  --ca CA | --ca-file FILE                 certificate authority
  --[no-]use-http-proxy                    allow/disallow HTTP proxy server usage (default: use, except for
                                           localhost)
  --package-fetch-poll-interval SECONDS    how frequently to fetch ALL packages (default: 0 [don't
                                           periodically fetch])
```

The old spew:

```
usage: dazl ls [-h] [--config CONFIG] [--max-connection-count COUNT] [--quiet-timeout SECONDS]
               [--use-acs-service] [--idle-timeout SECONDS] [--max-consequence-depth COUNT]
               [--package-ids PACKAGE_IDS_TYPE] [--server-host STRING] [--server-port PORT]
               [--oauth-client-id STRING] [--oauth-client-secret STRING] [--oauth-token STRING]
               [--oauth-refresh-token STRING] [--oauth-id-token STRING] [--oauth-token-uri STRING]
               [--oauth-redirect-uri STRING] [--oauth-auth-url STRING] [--oauth-ca-file STRING]
               [--oauth-audience STRING] [--oauth-legacy-username STRING] [--oauth-legacy-password STRING]
               [--parties PARTIES] [--url URL] [--ca-file PATH] [--cert-file PATH] [--cert-key-file PATH]
               [--verify-ssl VERIFY_SSL] [--poll-interval SECONDS] [--connect-timeout SECONDS]
               [--eager-package-fetch] [--enable-http-proxy] [--package-lookup-timeout SECONDS]
               [--application-name STRING] [--log-level LOG_LEVEL] [--max-event-block-size COUNT]
               [--max-command-batch-size COUNT] [--max-command-batch-timeout SECONDS]
               [--party-groups PARTIES] [--format FORMAT] [--template-filter TEMPLATE_FILTER] [--all]

optional arguments:
  -h, --help            show this help message and exit
  --format FORMAT, --fmt FORMAT, -F FORMAT
  --template-filter TEMPLATE_FILTER, -T TEMPLATE_FILTER
  --all, -A

Configuration Settings:
  --config CONFIG       path to a YAML config file

Overall Settings:
  --max-connection-count COUNT
                        Number of concurrent HTTP connections to have outstanding.
  --quiet-timeout SECONDS
                        Number of seconds to wait after the client "thinks" it's done to hang around for
  --use-acs-service     Use Active Contract Set service instead of reading from the Transaction event
                        stream
  --idle-timeout SECONDS
                        Maximum number of seconds of idle activity before automatically closing the client
  --max-consequence-depth COUNT
                        The maximum number of times to wait for all parties to arrive at the same
                        offsetbefore failing with an error
  --package-ids PACKAGE_IDS_TYPE
                        comma-separated list of package IDs to listen on
  --server-host STRING  Server listening host. Used for OAuth web application flow callbacks.
  --server-port PORT    Server listening port. Used for OAuth web application flow callbacks.
  --oauth-client-id STRING
                        OAuth client ID
  --oauth-client-secret STRING
                        OAuth client secret (implies web application flow or backend application flow)
  --oauth-token STRING  OAuth token
  --oauth-refresh-token STRING
                        OAuth refresh token
  --oauth-id-token STRING
                        OpenID token (JWT formatted)
  --oauth-token-uri STRING
                        OAuth token URL
  --oauth-redirect-uri STRING
                        OAuth redirect URI (implies web application flow)
  --oauth-auth-url STRING
                        OAuth auth URL (implies mobile application flow)
  --oauth-ca-file STRING
                        OAuth CA CA Bundle File
  --oauth-audience STRING
                        OAuth audience (Implies Client Credentials Flow)
  --oauth-legacy-username STRING
                        OAuth username (implies legacy application flow)
  --oauth-legacy-password STRING
                        OAuth password (implies legacy application flow)
  --parties PARTIES, -p PARTIES, --party PARTIES
                        comma-separated list of parties serviced by a participant node

Per-Party Settings:
  --url URL, -u URL, --participant-url URL
                        URL where the Ledger API is hosted
  --ca-file PATH        server certificate authority file
  --cert-file PATH      client certificate file
  --cert-key-file PATH  client certificate and key file
  --verify-ssl VERIFY_SSL
                        level of verification to use for SSL
  --poll-interval SECONDS
                        polling interval for receiving new events
  --connect-timeout SECONDS
                        number of seconds before giving up on a connection
  --eager-package-fetch
                        whether to fetch all packages on startup
  --enable-http-proxy   whether to allow the use of HTTP proxies
  --package-lookup-timeout SECONDS
                        number of seconds before giving up on a package
  --application-name STRING
                        The name that this application uses to identify itself to the ledger.
  --log-level LOG_LEVEL, -l LOG_LEVEL
                        logging level for events for this party
  --max-event-block-size COUNT
                        Maximum number of blocks to read in a single call.
  --max-command-batch-size COUNT
                        Maximum number of commands to batch up in a single transaction
  --max-command-batch-timeout SECONDS
                        Maximum number of seconds to wait before sending out a command
  --party-groups PARTIES
                        comma-separated list of broadcast parties
```

### DIFF
--- a/python/dazl/ledger/__init__.py
+++ b/python/dazl/ledger/__init__.py
@@ -1,9 +1,9 @@
 # Copyright (c) 2017-2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-
 """
 :mod:`dazl.ledger` — gRPC Ledger API / HTTP JSON API client
 ===========================================================
+.. autofunction:: connect
 
 This module contains the types needed to submit commands to and read events from a
 Daml `gRPC Ledger API <https://docs.daml.com/app-dev/ledger-api.html>`_ or
@@ -100,7 +100,11 @@ Other
     :members:
 
 """
+from logging import Logger
+from os import PathLike
+from typing import Collection, Optional, Union
 
+from ..prim import Party, TimeDeltaLike
 from .api_types import (
     ArchiveEvent,
     Boundary,
@@ -126,3 +130,209 @@ __all__ = [
     "ExerciseResponse",
     "PartyInfo",
 ]
+
+
+def connect(
+    url: Optional[str] = None,
+    host: Optional[str] = None,
+    port: Optional[int] = None,
+    scheme: Optional[str] = None,
+    read_as: Union[None, Party, Collection[Party]] = None,
+    act_as: Union[None, Party, Collection[Party]] = None,
+    admin: Optional[bool] = False,
+    ledger_id: Optional[str] = None,
+    application_name: Optional[str] = None,
+    oauth_token: Optional[str] = None,
+    oauth_token_file: Optional[str] = None,
+    ca: Optional[bytes] = None,
+    ca_file: Optional[PathLike] = None,
+    cert: Optional[bytes] = None,
+    cert_file: Optional[PathLike] = None,
+    cert_key: Optional[bytes] = None,
+    cert_key_file: Optional[PathLike] = None,
+    connect_timeout: Optional[TimeDeltaLike] = None,
+    use_http_proxy: bool = True,
+    logger: Optional[Logger] = None,
+    logger_name: Optional[str] = None,
+    log_level: Optional[str] = None,
+):
+    """
+    Create a connection from the supplied parameters.
+
+    The remote can be configured either by supplying a URL or by supplying a host (and optional
+    port and scheme). If none of ``url``, ``host``, ``port``, or ``scheme`` are supplied, then
+    environment variables are consulted; if no environment variables are specified either, then
+    default values are used.
+
+    The remote can be configured either by supplying a URL or by supplying a host (and optional
+    port and scheme). If none of ``url``, ``host``, ``port``, or ``scheme`` are supplied, then
+    environment variables are consulted; if no environment variables are specified either, then
+    default values are used.
+
+    When the URL scheme is ``http`` or ``https``, dazl will first attempt to connect assuming
+    the HTTP JSON API; if this fails, gRPC Ledger API is attempted.
+
+    If ``oauth_token`` is supplied and non-empty, then *token-based access* is used to connect
+    to the ledger. (Note that you cannot specify both ``oauth_token`` and ``oauth_token_file``.)
+
+    If other fields are specified, then *property-based access* is used. At least one of
+    ``read_as``, ``act_as``, or ``admin`` must be supplied. For the HTTP JSON API,
+    ``ledger_id`` MUST be supplied.
+
+    If _no_ fields are specified, this is an error unless environment variables supply an
+    alternate source of configuration.
+
+    +----------------------+----------------------+------------------------------------------------+
+    | function argument    | environment variable | default value                                  |
+    +======================+======================+================================================+
+    | **Connection config**                                                                        |
+    |                                                                                              |
+    | Specifying *any* of ``url``, ``host``, ``port``, or ``scheme`` causes *all* of these         |
+    | environment variables to be ignored                                                          |
+    +----------------------+----------------------+------------------------------------------------+
+    | ``url``              | DAML_LEDGER_URL      | localhost:6865                                 |
+    +----------------------+----------------------+------------------------------------------------+
+    | ``host``             | DAML_LEDGER_HOST     | localhost                                      |
+    +----------------------+----------------------+------------------------------------------------+
+    | ``port``             | DAML_LEDGER_PORT     | | 6865, unless ``scheme`` is specified:        |
+    |                      |                      | | • 80 for ``http``                            |
+    |                      |                      | | • 443 for ``https``                          |
+    |                      |                      | | • 6865 for ``grpc``                          |
+    +----------------------+----------------------+------------------------------------------------+
+    | ``scheme``           | DAML_LEDGER_SCHEME   | | ``https`` for port 443 or 8443               |
+    |                      |                      | | ``http`` for port 80, 7575 or 8080           |
+    |                      |                      | | ``grpc`` for port 6865                       |
+    |                      |                      | | ``https`` for all other ports                |
+    +----------------------+----------------------+------------------------------------------------+
+    | **Access config**                                                                            |
+    |                                                                                              |
+    | Specifying *any* of ``act_as``, ``read_as``, ``admin``, ``ledger_id``, ``application_name``, |
+    | ``oauth_token``, or ``oauth_token_file`` causes *all* of these environment variables to be   |
+    | ignored                                                                                      |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``act_as``           | ``DAML_LEDGER_ACT_AS``                                                |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``read_as``          | ``DAML_LEDGER_READ_AS``                                               |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``ledger_id``        | ``DAML_LEDGER_ID``                                                    |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``application_name`` | ``DAML_LEDGER_APPLICATION_NAME``                                      |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``oauth_token``      | ``DAML_LEDGER_OAUTH_TOKEN``                                           |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``oauth_token_file`` | ``DAML_LEDGER_OAUTH_TOKEN_FILE``                                      |
+    +----------------------+-----------------------------------------------------------------------+
+
+    .. note::
+
+        When connecting to the gRPC Ledger API, note that
+        `gRPC environment variables
+        <https://grpc.github.io/grpc/cpp/md_doc_environment_variables.html>`_ are also respected.
+        You can configure the gRPC client in much more detail with these flags.
+
+    :param url:
+        The URL to connect to. Can be used as an alternative to supplying ``host``, ``port``,
+        and ``scheme`` as individual values. Can alternatively be supplied via the environment
+        variable ``DAML_LEDGER_URL``.
+    :param host:
+        The host to connect to. Can be used as an alternative to supplying ``url`` as a combined
+        value. Can alternatively be supplied via the environment variable ``DAML_LEDGER_HOST``.
+    :param port:
+        The port to connect to. Can be used as an alternative to supplying ``url`` as a combined
+        value. Can alternatively be supplied via the environment variable ``DAML_LEDGER_PORT``.
+    :param scheme:
+        The scheme to connect to. Can be used as an alternative to supplying ``url`` as a
+        combined value. Can alternatively be supplied via the environment variable
+        ``DAML_LEDGER_SCHEME``.
+    :param connect_timeout:
+        Length of time to wait before giving up connecting to the remote and declaring an error.
+        The default value is 30 seconds.
+    :param use_http_proxy:
+        ``True`` to use an HTTP(S) proxy server if configured; ``False`` to avoid using any
+        configured server. If unspecified and the host is ``localhost``, the proxy server is
+        avoided; otherwise the proxy server is used.
+    :param ca:
+        A certificate authority to use to validate the server's certificate. If not supplied,
+        the operating system's default trust store is used. Cannot be specified with
+        ``ca_file``.
+    :param ca_file:
+        A file containing the certificate authority to use to validate the server's certificate.
+        If not supplied, the operating system's default trust store is used. Cannot be specified
+        with ``ca``.
+    :param cert:
+        A client-side certificate to be used when connecting to a server that requires mutual
+        TLS. Cannot be specified with ``cert_file``.
+    :param cert_file:
+        A file containing the client-side certificate to be used when connecting to a server
+        that requires mutual TLS. Cannot be specified with ``cert``.
+    :param cert_key:
+        A client-side private key to be used when connecting to a server that requires mutual
+        TLS. Cannot be specified with ``cert_key_file``.
+    :param cert_key_file:
+        A client-side private key to be used when connecting to a server that requires mutual
+        TLS. Cannot be specified with ``cert_key``.
+    :param read_as:
+        A party or set of parties on whose behalf (in addition to all parties listed in
+        ``act_as``) contracts can be retrieved. Cannot be specified if ``oauth_token`` or
+        ``oauth_token_file`` is specified.
+    :param act_as:
+        A party or set of parties on whose behalf commands should be executed. Parties here are
+        also implicitly granted ``read_as`` access as well. Cannot be specified if
+        ``oauth_token`` or ``oauth_token_file`` is specified.
+    :param admin:
+        HTTP JSON API only: allow admin endpoints to be used. This flag is ignored when
+        connecting to gRPC Ledger API implementations. Cannot be specified if ``oauth_token`` or
+        ``oauth_token_file`` is specified.
+    :param ledger_id:
+        The ledger ID to connect to. For the HTTP JSON API, this value is required. For the gRPC
+        Ledger API, if this value is _not_ supplied, its value will be retrieved from the
+        server. Cannot be specified if ``oauth_token`` or ``oauth_token_file`` is specified.
+    :param application_name:
+        A string that identifies this application. This is used for tracing purposes on the
+        server-side. Cannot be specified if ``oauth_token`` or ``oauth_token_file`` is
+        specified.
+    :param oauth_token:
+        The OAuth bearer token to be used on all requests. If specified, no other access
+        parameters can be specified.
+    :param oauth_token_file:
+        A file that contains the OAuth bearer token to be used on all requests. If specified, no
+        other access parameters can be specified.
+    :param logger:
+        The logger to use for connections created from this configuration. If not supplied, a
+        logger will be created.
+    :param logger_name:
+        The name of the logger. Only used if ``logger`` is not provided.
+    :param log_level:
+        The logging level for the logger. The default is ``warn``. Only used if ``logger`` is
+        not provided.
+    """
+    # TODO: Ideally we could _define_ the function as taking all of these parameters, but implement
+    #  it with kwargs passing; unfortunately Sphinx doesn't seem to allow this.
+    from .config import Config
+
+    _ = Config.create(
+        url=url,
+        host=host,
+        port=port,
+        scheme=scheme,
+        read_as=read_as,
+        act_as=act_as,
+        admin=admin,
+        ledger_id=ledger_id,
+        application_name=application_name,
+        oauth_token=oauth_token,
+        oauth_token_file=oauth_token_file,
+        ca=ca,
+        ca_file=ca_file,
+        cert=cert,
+        cert_file=cert_file,
+        cert_key=cert_key,
+        cert_key_file=cert_key_file,
+        connect_timeout=connect_timeout,
+        use_http_proxy=use_http_proxy,
+        logger=logger,
+        logger_name=logger_name,
+        log_level=log_level,
+    )
+    # TODO: Implement when the v8 connection code fully lands
+    raise NotImplementedError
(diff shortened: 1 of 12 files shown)


python/dazl/ledger/__init__.py
```diff
@@ -24,3 +34,209 @@
...
+    :param logger_name:
+        The name of the logger. Only used if ``logger`` is not provided.
+    :param log_level:
+        The logging level for the logger. The default is ``warn``. Only used if ``logger`` is
+        not provided.
+    """
+    # TODO: Ideally we could _define_ the function as taking all of these parameters, but implement
+    #  it with kwargs passing; unfortunately Sphinx doesn't seem to allow this.
+    from .config import Config
+
```
lima-da, line 216: Looking at the signatures and the docs seems like the difficult part is to handle the precedence of different configuration combinations because most fields are optional and flat?  thinking if there is a way to group these combinations to separate classes that will result in general "ledger connection context", e.g.
for access: (read_as, act_as, admin) | (oauth_token) | (oauth_token_file)
for connection: json_api_conn | grpc_conn
?

da-tanabe: I'm trying to have my cake and eat it too; I'd like client code to look as simple as possible because most times people simply won't be using lots of parameters:

```py
# the IDE can help you here too because all the parameters are right there
conn = dazl.connect(url="https://ledger.daml.app", oauth_token_file="token.txt")
```

What you're proposing is still doable in this PR too:

```py
from dazl.ledger import Connection, Config
from dazl.ledger.config import create_access, create_url, SimpleURLConfig, TokenBasedAccessConfig

conn = Connection(Config(url=create_url("https://ledger.daml.app"), access=create_access(oauth_token_file="token.txt"))

# or alternatively
conn = Connection(Config(url=SimpleURLConfig("https://ledger.daml.app"), access=TokenBasedAccessConfig(oauth_token_file="token.txt"))
```

I opted for convenience layers to keep it down to one line, but _ultimately_ implement it the way you suggested where different classes implement the different cases.

python/dazl/ledger/config/access.py
```diff
@@ -0,0 +1,491 @@
...
+    elif oauth_token:
+        return TokenBasedAccessConfig(oauth_token)
+    else:
+        return PropertyBasedAccessConfig(
+            read_as=read_as,
+            act_as=act_as,
+            admin=admin,
+            ledger_id=ledger_id,
+            application_name=application_name,
+        )
```
lima-da, line 107: So the precedence is `oauth_token_file ` > `oauth_token` > `readAs, actAs, admin` ? then maybe in the docs we should say xxx will be ignored if xxx is specified first, instead of "Cannot be specified" ?

da-tanabe: No there's no precedence between the variables—if conflicting values are specified, then you get a `ConfigError`; the only precedence rule is that if you pass access parameters, the environment variables as a whole are now ignored. So in other words:

```bash
DAML_LEDGER_OAUTH_TOKEN_FILE=token.txt python -c "import dazl; dazl.connect(act_as='Alice')" 
```

will ignore the environment variable, because `act_as` was specified.

python/dazl/ledger/config/access.py
```diff
@@ -0,0 +1,491 @@
...
+
+    @property
+    def application_name(self) -> str:
+        return self._application_name
+
+
+class TokenFileBasedAccessConfig(TokenBasedAccessConfig):
+    def __init__(self, oauth_token_file: str):
+        # TODO: Update the token when the contents of this file change.
+        super().__init__(Path(oauth_token_file).read_text())
```
lima-da, line 253: If this were supported, does that means the auth pattern for refresh token would be a side container/process will be needed for writing new tokens? (similar to how json-api works)

python/dazl/ledger/config/access.py
```diff
@@ -0,0 +1,491 @@
...
+        self._admin = bool(admin)
+        self._ledger_id = ledger_id
+        self._application_name = application_name or "dazl-client"
+
+    @property
+    def token(self):
+        """
+        Produces a token without signing, utilizing our parameters.
+        """
+        return encode_unsigned_token(
```
lima-da, line 302: Ah, so property based mode are technically no-auth mode?

python/dazl/ledger/config/access.py
da-tanabe: Yup that _could_ be an option. It would also be possible to implement `AccessConfig` manually and just do whatever needs to be done (because now it's an explicit interface inside the config system that makes less assumptions about the object).

But _ultimately_ my preferred approach would be to implement the standard oauth refresh flow here and expand the scope of what `TokenBasedAccessConfig` could handle (refresh URLs, refresh tokens, etc). Honestly the biggest blocker here is just building up a proper test suite in this repo where dazl tries to connect to an authenticated ledger that supports refreshing—it's a lot of mock machinery to set up.

python/dazl/ledger/config/access.py
da-tanabe: yup—either token-based or property-based and they can't really be mixed.

python/dazl/ledger/config/access.py
mschaef-da: Does the config error illustrate what needs to be fixed?

python/dazl/ledger/__init__.py
lima-da: ah I see. as long as explicit config is possible. OK,  make sense that dazl is meant to serve things less verbose in the most use case.👍

python/dazl/ledger/__init__.py
```diff
@@ -24,3 +34,209 @@
...
+    environment variables are consulted; if no environment variables are specified either, then
+    default values are used.
+
+    The remote can be configured either by supplying a URL or by supplying a host (and optional
+    port and scheme). If none of ``url``, ``host``, ``port``, or ``scheme`` are supplied, then
+    environment variables are consulted; if no environment variables are specified either, then
+    default values are used.
+
+    When the URL scheme is ``http`` or ``https``, dazl will first attempt to connect assuming
+    the HTTP JSON API; if this fails, gRPC Ledger API is attempted.
```
mschaef-da, line 77: How do the semantics change between the two back ends? Does the data format produced and accepted change at all? IIRC, the JSON API doesn't support reading previously archived contracts either. It seems like there's a significant opportunity for 'spooky action at a distance' in that the way a ledger connection is setup can have a significant impact on the way that connection behaves that is not obvious at the point the connection is used.

python/dazl/ledger/config/access.py
```diff
@@ -0,0 +1,491 @@
...
+
+    is_property_based = read_as or act_as or admin or ledger_id or application_name
+    if not is_property_based and not oauth_token and not oauth_token_file:
+        raise ConfigError("no oauth token access or read_as/act_as/admin was specified")
+
+    # how do they configure thee? let me count the ways...
+    if sum(map(int, (is_property_based, oauth_token, oauth_token_file))):
+        raise ConfigError(
+            "must specify ONE of read_as/act_as/admin, oauth_token, or oauth_token_file"
+        )
```
da-tanabe, line 94: @mschaef-da yup here's the error string (GH makes it annoying to cross-reference in the same PR)…

python/dazl/ledger/config/access.py
da-tanabe: yup commented elsewhere.

python/dazl/ledger/__init__.py
```diff
@@ -24,3 +34,209 @@
...
+    If _no_ fields are specified, this is an error unless environment variables supply an
+    alternate source of configuration.
+
+    +----------------------+----------------------+------------------------------------------------+
+    | function argument    | environment variable | default value                                  |
+    +======================+======================+================================================+
+    | **Connection config**                                                                        |
+    |                                                                                              |
+    | Specifying *any* of ``url``, ``host``, ``port``, or ``scheme`` causes *all* of these         |
+    | environment variables to be ignored                                                          |
```
mschaef-da, line 95: I haven't gotten far enough to see if you've done this yet, but given the complexity of this config mechanism, there really should be a way to request a log message indicating exactly what configuration values are in use and which of the various mechanisms have been given authority under the precedence scheme. (Very glad to see it in the docs, though.)

python/dazl/ledger/__init__.py
```diff
@@ -24,3 +34,209 @@
...
+    +----------------------+----------------------+------------------------------------------------+
+    | **Access config**                                                                            |
+    |                                                                                              |
+    | Specifying *any* of ``act_as``, ``read_as``, ``admin``, ``ledger_id``, ``application_name``, |
+    | ``oauth_token``, or ``oauth_token_file`` causes *all* of these environment variables to be   |
+    | ignored                                                                                      |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``act_as``           | ``DAML_LEDGER_ACT_AS``                                                |
+    +----------------------+-----------------------------------------------------------------------+
+    | ``read_as``          | ``DAML_LEDGER_READ_AS``                                               |
```
mschaef-da, line 119: It would be good to explicitly discuss the way that multiple parties can be specified for these when they take their value from an environment variable.

python/dazl/ledger/__init__.py
```diff
@@ -24,3 +34,209 @@
...
+    :param host:
+        The host to connect to. Can be used as an alternative to supplying ``url`` as a combined
+        value. Can alternatively be supplied via the environment variable ``DAML_LEDGER_HOST``.
+    :param port:
+        The port to connect to. Can be used as an alternative to supplying ``url`` as a combined
+        value. Can alternatively be supplied via the environment variable ``DAML_LEDGER_PORT``.
+    :param scheme:
+        The scheme to connect to. Can be used as an alternative to supplying ``url`` as a
+        combined value. Can alternatively be supplied via the environment variable
+        ``DAML_LEDGER_SCHEME``.
```
mschaef-da, line 150: This text _"Can alternatively be supplied via the environment variable ``DAML_LEDGER_URL``."_ should probably also carry the caveat above regarding ALL environment variables being ignored if any of these are specified. Otherwise if all you read is the parameter documentation (tl;dr ftw!) you might miss the fact that the env var can be ignored.

python/dazl/ledger/__init__.py
mschaef-da: Agreed with Li here... this is not necessarily easy logic to understand.  I think it can be addressed with a few logging facilities and documentation.

python/dazl/ledger/config/argv.py
```diff
@@ -0,0 +1,386 @@
...
+if TYPE_CHECKING:
+    from argparse import _ArgumentGroup
+
+__all__ = ["configure_parser", "EXAMPLES"]
+
+
+# Implementation notes:
+#
+# A common trick employed throughout this file is to add an "argument" to a parser that is not
+# directly used. Calling ``add_arguments(...)`` with an argumen
```
mschaef-da, line 24: Looks like there is more you wanted to say. (And this talks about the trick without explaining what is trying to accomplish and why that is a worthwhile goal.)

python/dazl/ledger/config/access.py
lima-da: Yeah, agree on the "standard oauth refresh flow". This would open the door for patterns like auth-middleware.
and yea.. 

about test suites maybe it's easier to do integration tests that spin a ledger container and let the test generates keypair just for testing the auth flow?

python/dazl/ledger/config/argv.py
```diff
@@ -0,0 +1,386 @@
...
+
+        super().__init__(prog, *args, **kwargs)
+
+
+def configure_parser(parser: ArgumentParser, add_usage: bool = True) -> None:
+    """
+    Add dazl configuration options to an existing :class:`ArgumentParser`.
+
+    dazl exposes a variety of configuration options that could be tedious to duplicate in your own
+    application. To use this in your own application:
```
mschaef-da, line 54: This is a nice idea, but runs the risk of both present and future collisions between user application arguments and dazl arguments. (It's not hard to imagine that somebody wants to write a user app that accepts a `--port` option specific to their application and distinct from DAZL's port option, etc.)

python/dazl/ledger/config/argv.py
```diff
@@ -0,0 +1,386 @@
...
+    group_p.add_argument(
+        "--admin",
+        action="store_const",
+        const=True,
+        help="Add admin flag (HTTP JSON API only)",
+    )
+    group_p.add_argument(
+        "--ledger-id",
+        "-l",
+        help="ledger ID (required on HTTP JSON API)",
```
mschaef-da, line 109: Are there warnings thrown when you say `--admin` to a gRPC setup or don't say `--ledger-id` to a HTTP JSON setup?

python/dazl/ledger/config/argv.py
```diff
@@ -0,0 +1,386 @@
...
+    )
+    group_p.add_argument(
+        "--application-name",
+        metavar="NAME",
+        help="application name",
```
mschaef-da, line 114: Maybe more descriptive text. (A ref to the DAML docs that describe what this is and is for?)

python/dazl/ledger/config/access.py
da-tanabe: > about test suites maybe it's easier to do integration tests that spin a ledger container and let the test generates keypair just for testing the auth flow?

That'd be step 1, but if we implemented refresh flow support I'd want to make sure that specifically works as well.

python/dazl/ledger/config/url.py
```diff
@@ -0,0 +1,211 @@
...
+        url = build_url(host, port, scheme)
+
+    if use_http_proxy is None:
+        use_http_proxy = not is_localhost(urlparse(url).hostname)
+
+    return SimpleURLConfig(
+        url=url,
+        connect_timeout=to_timedelta(connect_timeout) if connect_timeout is not None else None,
+        use_http_proxy=use_http_proxy,
+    )
```
mschaef-da, line 75: Debug logging through a lot of this would be useful for people to understand why things are happening the way they are.

python/docs/migrating.rst
```diff
@@ -5,6 +5,50 @@
...
+   - ``--max-connection-batch-size``
+   - ``--max-connection-count``
+   - ``--max-consequence-depth``
+   - ``--max-event-block-size``
+   - ``--poll-interval``
+   - ``--quiet-count``
+   - ``--use-acs-service``
+
+
 Migrating from dazl v6 from v7
```
lima-da, line 52: Other than command-line change, any other migration notes for underneath the actual client configuration for direct python package use that worth mention here? 
mostly for our own benefit, most of our dazl usage are direct python object calls

python/dazl/ledger/__init__.py
da-tanabe: A design goal is to actually make them look and feel as much the same as possible. Although the gRPC Ledger API makes it possible to return archive events from the past, `dazl` doesn't really expose them any more unless you explicitly said "read from the beginning", which is no longer an option in dazl 8.

I also dropped fields from the read-side (`event_id`, `witness_parties`) that have no analog in the HTTP JSON API, as they're not really used; but in return I _added_ fields (`signatories`, `observers`, etc) that are present in both, as `dazl` was a bit behind the Ledger API.

python/dazl/ledger/__init__.py
da-tanabe: Good point—will add to a subsequent PR.

python/dazl/ledger/__init__.py
da-tanabe: Agreed…will change.

python/dazl/ledger/config/argv.py
da-tanabe: oof hahaha yep…not sure what happened here.

python/dazl/ledger/config/argv.py
da-tanabe: Honestly I would have wanted to drop it altogether, but people actually use this stuff!

My preferred approach would be for people to design their own command lines and build up a config manually (which I think is the only way to ultimately avoid conflicts in parameters); but short of that, this is _already_ opt-in behavior so by calling this, you as the app author have already made the decision to trade off flexibility of your own parameters for convenience of not having to re-specify them.

python/dazl/ledger/config/argv.py
da-tanabe: On a gRPC setup, `--admin` is harmless as it simply has no effect; `--ledger-id` will annoyingly crash at the time of connection.

From a user point of view, the exception will always originate from the one line:

```py
conn = dazl.connect(url="localhost:7575", act_as="Alice")
# ^ will throw because ledger ID is missing
```

python/dazl/ledger/config/argv.py
da-tanabe: yep shall do

python/docs/migrating.rst
da-tanabe: Yup for sure—I'm adding stuff to this section in chunks as more of the new API gets merged in (see #108).

python/dazl/ledger/config/argv.py
da-tanabe: Bleah on second look the docs aren't that great (https://github.com/digital-asset/daml/blob/691edeacf2cf2d480ad56f70d972f4ac4ee154ac/ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto#L31-L34); also I'm trying to keep these help strings as short as possible. This is documented in more detail in the `connect(...)` function though, where there is a little more space to elaborate.

lima-da (approved):
Overall looks very nice, no blocker to merge this in first 🎉

python/dazl/ledger/__init__.py
da-tanabe: Revised substantially to be clearer.

python/dazl/ledger/config/url.py
da-tanabe: Agreed and added.

PR #179, opened 2021-03-16
python: Documentation fixes for dazl.ledger.api_types.

da-tanabe:
Some minor documentation fixes here and there, mostly in `dazl.ledger.api_types`.

### DIFF
--- a/python/dazl/ledger/__init__.py
+++ b/python/dazl/ledger/__init__.py
@@ -1,8 +1,97 @@
 # Copyright (c) 2017-2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+"""
+:mod:`dazl.ledger` — gRPC Ledger API / HTTP JSON API client
+===========================================================
+
+This module contains the types needed to submit commands to and read events from a
+Daml `gRPC Ledger API <https://docs.daml.com/app-dev/ledger-api.html>`_ or
+`HTTP JSON API <https://docs.daml.com/json-api/index.html>`_.
+
+Example:
+
++-----------------------------------+--------------------------------------------------------------+
+| Commands                                                                                         |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`Command`                  | abstract base class of all commands                          |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`CreateCommand`            | create a contract                                            |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`CreateAndExerciseCommand` | create a contract and immediately exercise a choice in a     |
+|                                   | single transaction                                           |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ExerciseCommand`          | exercises a choice on a contract identified by its contract  |
+|                                   | ID                                                           |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ExerciseByKeyCommand`     | exercises a choice on a contract identified by its contract  |
+|                                   | key                                                          |
++-----------------------------------+--------------------------------------------------------------+
+| Events and Responses                                                                             |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`CreateEvent`              | event raised when a contract is created, either as part of   |
+|                                   | the active contract set or on the transaction stream         |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ArchiveEvent`             | event raised when a contract is archived on the transaction  |
+|                                   | stream                                                       |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ExerciseResponse`         | the response from an exercise, including the choice's return |
+|                                   | value (if any)                                               |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`Boundary`                 | indicates a point where an event stream can be subsequently  |
+|                                   | resumed                                                      |
++-----------------------------------+--------------------------------------------------------------+
+| Other                                                                                            |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`PartyInfo`                | metadata about a party                                       |
++-----------------------------------+--------------------------------------------------------------+
+
+Command types
+-------------
+
+.. autoclass:: Command
+
+.. autoclass:: CreateCommand
+   :members:
+
+.. autoclass:: CreateAndExerciseCommand
+   :members:
+
+.. autoclass:: ExerciseCommand
+   :members:
+
+.. autoclass:: ExerciseByKeyCommand
+   :members:
+
+Event types
+-----------
+
+.. autoclass:: CreateEvent
+   :members:
+
+.. autoclass:: ArchiveEvent
+   :members:
+
+Other read-side types
+---------------------
+
+.. autoclass:: Boundary
+   :members:
+
+.. autoclass:: ExerciseResponse
+   :members:
+
+Other
+-----
+
+.. autoclass:: PartyInfo
+    :members:
+
+"""
+
 from .api_types import (
     ArchiveEvent,
+    Boundary,
     Command,
     CreateAndExerciseCommand,
     CreateCommand,
@@ -15,6 +104,7 @@ from .api_types import (
 
 __all__ = [
     "ArchiveEvent",
+    "Boundary",
     "Command",
     "CreateAndExerciseCommand",
     "CreateCommand",
--- a/python/dazl/ledger/__init__.py
+++ b/python/dazl/ledger/__init__.py
@@ -18,13 +18,13 @@ Example:
 +-----------------------------------+--------------------------------------------------------------+
 | :class:`CreateCommand`            | create a contract                                            |
 +-----------------------------------+--------------------------------------------------------------+
-| :class:`CreateAndExerciseCommand` | create a contract and immediately exercise a choice in a     |
-|                                   | single transaction                                           |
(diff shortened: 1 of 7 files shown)


python/dazl/ledger/__init__.py
```diff
@@ -1,8 +1,97 @@
...
+This module contains the types needed to submit commands to and read events from a
+Daml `gRPC Ledger API <https://docs.daml.com/app-dev/ledger-api.html>`_ or
+`HTTP JSON API <https://docs.daml.com/json-api/index.html>`_.
+
+Example:
+
++-----------------------------------+--------------------------------------------------------------+
+| Commands                                                                                         |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`Command`                  | abstract base class of all commands                          |
```
mschaef-da, line 17: any reason the descriptions here are not capitalized and lack punctuation

python/dazl/ledger/__init__.py
```diff
@@ -1,8 +1,97 @@
...
++-----------------------------------+--------------------------------------------------------------+
+| :class:`CreateCommand`            | create a contract                                            |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`CreateAndExerciseCommand` | create a contract and immediately exercise a choice in a     |
```
mschaef-da, line 21: exercise a choice *on the newly created contract*

python/dazl/ledger/__init__.py
```diff
@@ -1,8 +1,97 @@
...
+|                                   | single transaction                                           |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ExerciseCommand`          | exercises a choice on a contract identified by its contract  |
```
mschaef-da, line 24: s/exercises/exercise/

(To be consistent with create rather than creates)

python/dazl/ledger/__init__.py
```diff
@@ -1,8 +1,97 @@
...
++-----------------------------------+--------------------------------------------------------------+
+| Events and Responses                                                                             |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`CreateEvent`              | event raised when a contract is created, either as part of   |
+|                                   | the active contract set or on the transaction stream         |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ArchiveEvent`             | event raised when a contract is archived on the transaction  |
+|                                   | stream                                                       |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ExerciseResponse`         | the response from an exercise, including the choice's return |
```
mschaef-da, line 38: "event raised when a contract choice is exercised, including...."

python/dazl/ledger/__init__.py
```diff
@@ -1,8 +1,97 @@
...
+Event types
+-----------
+
+.. autoclass:: CreateEvent
+   :members:
+
+.. autoclass:: ArchiveEvent
+   :members:
+
+Other read-side types
```
mschaef-da, line 75: I've always liked the distinction between 'read' and 'write' sides of the API, but the write side types might should be denoted as such.

python/dazl/ledger/__init__.py
mschaef-da: This might also ought to be an `ExerciseResponseEvent` class, if that can be done in a backward compat way. (So write-side commands are `Command` and read-side events are `Event`)

python/dazl/ledger/__init__.py
```diff
@@ -1,8 +1,97 @@
...
+| :class:`CreateEvent`              | event raised when a contract is created, either as part of   |
+|                                   | the active contract set or on the transaction stream         |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ArchiveEvent`             | event raised when a contract is archived on the transaction  |
+|                                   | stream                                                       |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`ExerciseResponse`         | the response from an exercise, including the choice's return |
+|                                   | value (if any)                                               |
++-----------------------------------+--------------------------------------------------------------+
+| :class:`Boundary`                 | indicates a point where an event stream can be subsequently  |
```
mschaef-da, line 41: `BoundaryEvent`?

python/dazl/ledger/__init__.py
da-tanabe: I copied the style of the table at the top of https://docs.python.org/3/library/collections.html ; I'm not a _huge_ fan of this treatment, but when in Rome I guess…

python/dazl/ledger/__init__.py
da-tanabe: Another thing I agonized over!

I went back and forth on this; I opted _not_ to call it an `Event` in order to align with the more narrow definition of events over the Ledger API and in the TypeScript API (`CreateEvent`/`ArchiveEvent`); _boundaries_ aren't really events as they are mere indications of places where streams could be resumed if you wanted to. They aren't necessarily repeatable because the HTTP JSON API doesn't necessarily promise it will give you the same boundaries if you requested a stream again later.

From an API standpoint you won't see them much, as I expect most clients will only be concerned with Ledger API events and not necessarily offset resuming, especially in the absence of a backing store where such offset could information could be stored:

```py
# creates only (sweep)
conn = dazl.connect(...)
async with conn.stream("Some:Template") as stream:
    async for cev in stream.creates():
        print(f"{cev.contract_id} -> {cev.payload}")
```

```py
# creates and archives--in other words, all "events" as you would see them documented elsewhere
conn = dazl.connect(...)
async with conn.stream("Some:Template") as stream:
    async for cev in stream.events():
        if isinstance(ev, CreateEvent):
             print(f"CREATE {ev.contract_id} -> {ev.payload}")
        elif isinstance(ev, ArchiveEvent):
             print(f"ARCHIVE {ev.contract_id} -> {ev.payload}")
        else:
             # no other type of Ledger API event
```

```py
# EVERYTHING, including things that aren't "events"
conn = dazl.connect(...)
async with conn.stream("Some:Template") as stream:
    async for i in stream.items():
        if isinstance(i, CreateEvent):
             print(f"CREATE {i.contract_id} -> {i.payload}")
        elif isinstance(i, ArchiveEvent):
             print(f"ARCHIVE {i.contract_id} -> {i.payload}")
        elif isinstance(i, Boundary):
             print("BOUNDARY offset {i.offset}")
        else:
             # no other kind of object in the stream
```

python/dazl/ledger/__init__.py
da-tanabe: See the comment below; I'm (now) trying to avoid calling things `Event`'s that are not on the Ledger API, and essentially would like to one day declare `Event` the same way it's declared in the TypeScript bindings (essentially `CreateEvent | ArchiveEvent`: https://docs.daml.com/app-dev/bindings-ts/daml-ledger/modules/_index_.html#event

Some names, like `ExerciseResponse`, are actually stolen from the TypeScript library as well. :-)

There has been some talk of `ArchiveEvent`'s gaining the same set of data as a `CreateEvent`, and if this happens, I'd even more so like to give the `Event` suffix special treatment as "a thing on the ledger that represents a state change of a _specific_ contract", as opposed to `dazl`'s previous looser treatment of it as anything that happened on the read side.

Arguably the introduction of read _streams_ already breaks the naming pattern of calling all read-side objects events (because a stream is a sequence of events), so from that perspective as well, I think tightening up the definitions of these suffices make sense.

python/dazl/ledger/__init__.py
da-tanabe: Oof…fixed.

python/dazl/ledger/__init__.py
da-tanabe: fixed

python/dazl/ledger/__init__.py
da-tanabe: Added some clarifying text. As I was doing that, I reminded myself that `ExerciseResponse` _contains_ events as well (just like a stream does); so even more so being more restrictive about what is and isn't an `Event` makes sense, I think:

```py
class ExerciseResponse:
    result: Optional[Any]
    events: Sequence[Union[CreateEvent, ArchiveEvent]]
```

For historical reasons I'm not ready to pull the trigger on `Event = Union[CreateEvent, ArchiveEvent]` even though it should probably read that way.

mschaef-da: Makes sense to me.

python/dazl/ledger/__init__.py
mschaef-da: I like that. (Not totally convinced on the use of stream methods as a filter specifier, but the whole thing is generally such a big improvement regardless.)

mschaef-da (approved):

python/dazl/ledger/__init__.py
da-tanabe: Changed.

python/dazl/ledger/__init__.py
da-tanabe: Yep that was a tough design decision too; it mostly ended up this way because the server _will_ send all this data down anyway (there's no way to ask the ledger for "just creates"), so the filter is client-side. I prototyped a version where `stream` had a bunch of parameters but I ended up just having to pass all that state down into the created stream _anyway_, not to mention trying to figure out how to make mypy happy.

At least with the functions on the stream, the typing rules are straightforward; otherwise I'd need a type parameter on the stream—and this is precisely where I punched out of trying to add them as parameters to the stream call, because `mypy` isn't _nearly_ powerful enough to express this. :-/

PR #185, opened 2021-03-19
Detach ACS view read and write operations.

mschaef-da:
This is not tested, but it should serve to illustrate the concept I'm advocating in your toplevel PR.

### DIFF
--- a/python/dazl/ledgerutil/acs.py
+++ b/python/dazl/ledgerutil/acs.py
@@ -2,7 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
 
-from collections.abc import MutableMapping
+import types
+
+from collections.abc import Mapping
 from operator import delitem, setitem
 from typing import Dict, Iterator, Optional, TypeVar, Union
 
@@ -16,8 +18,7 @@ from ..query import ContractMatch, is_match
 
 T = TypeVar("T")
 
-
-class ACS(MutableMapping):
+class ACS:
     """
     An accumulation of active contracts up to a certain point in time over a provided stream.
 
@@ -53,7 +54,7 @@ class ACS(MutableMapping):
         class MyService:
             def __init__(self, conn):
                 self._conn = conn
-                self._widgets = ACS.empty()
+                self._widgets = ACS.empty("WidgetMod:Widget")
 
             async def run(self):
                 async with self._conn.stream("WidgetMod:Widget") as stream:
@@ -61,44 +62,44 @@ class ACS(MutableMapping):
                     await stream.run()
 
             def widget_json(self, widget_id: str):
-                return self._widgets.first_payload({"id": widget_id})
+                return self._widgets.snapshot().first_payload({"id": widget_id})
 
     """
 
     @classmethod
-    def empty(cls) -> ACS:
-        return ACS()
+    def empty(cls, template_id: Union[str, TypeConName]) -> ACS:
+        return ACS(None, template_id)
 
     @classmethod
     def from_stream(cls, stream: QueryStream, template_id: Union[str, TypeConName]):
-        acs = ACS()
-        stream.on_create(template_id, lambda e: setitem(acs._contracts, e.contract_id, e.payload))
-        stream.on_archive(template_id, lambda e: delitem(acs._contracts, e.contract_id))
+        acs = ACS(0, template_id)
+        stream.on_create(template_id, lambda e: acs._set_item(e.contract_id, e.payload))
+        stream.on_archive(template_id, lambda e: acs._del_item(e.contract_id))
         return acs
 
-    def __init__(self):
+    def __init__(self, serial_number: Optional[int], template_id: Union[str, TypeConName]):
         self._contracts = {}  # type: Dict[ContractId, ContractData]
+        self._serial_number = serial_number
+        self._template_id = template_id
 
-    def __contains__(self, key: ContractId) -> bool:
-        return key in self._contracts
+    def _notice_change(self):
+        if self._serial_number is not None:
+            self._serial_number = self._serial_number + 1
 
-    def __getitem__(self, key: ContractId) -> ContractData:
-        return self._contracts[key]
+    def _set_item(self, key: ContractId, payload: ContractData) -> None:
+        self._notice_change()
+        setitem(self._contracts, key, payload)
 
-    def __setitem__(self, key: ContractId, payload: ContractData) -> None:
-        self._contracts[key] = payload
+    def _del_item(self, key: ContractId) -> None:
+        self._notice_change()
+        delitem(self._contracts, key)
 
-    def __delitem__(self, key: ContractId) -> None:
-        del self._contracts[key]
-
-    def __iter__(self) -> Iterator[ContractId]:
-        return iter(self._contracts)
-
-    def __len__(self) -> int:
-        return len(self._contracts)
-
-    def clear(self) -> None:
-        self._contracts.clear()
+    def snapshot(self) -> Mapping[ContractId, ContractData]:
+        """
+        Create a snapshot of the _current_ state of the ACS. Useful if you need a consistent view
+        of the ACS at a point in time.
+        """
+        return types.MappingProxyType(dict(self._contracts))
 
     def first_payload(self, predicate: ContractMatch) -> Optional[ContractData]:
         for value in self._contracts.values():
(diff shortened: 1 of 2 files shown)


python/dazl/ledgerutil/acs.py
```diff
@@ -17,7 +17,7 @@
 T = TypeVar("T")
 
 
-class ACS(MutableMapping):
+class ACS:
```
mschaef-da, line 20: This is key... the `ACS` itself is no longer a mapping - that task is moved into a separate snapshot object that the `ACS` can produce on request. The snapshot then represents a stable state of the ACS at the time the snapshot was taken.

python/dazl/ledgerutil/acs.py
```diff
@@ -65,19 +65,49 @@ def widget_json(self, widget_id: str):
...
         return acs
 
-    def __init__(self):
+    def __init__(self, template_id):
         self._contracts = {}  # type: Dict[ContractId, ContractData]
+        self._serial_number = 0
+        self._template_id = template_id
+
+    def _expect_serial_number(self, expected):
+        if self._serial_number != expected:
```
mschaef-da, line 81: Each update to the internal ACS maintained by an instance of `ACS` also updates a serial number within that instance. What this does is open up the possibility of a snapshot that can notice when the underlying ACS has changed on ledger and signal an error. The user of this ACS implementation then has a choice - `snapshot` that will return an ordinary snapshot and `strict_snapshot` that will attempt to detect ACS changes that occur out from under you.

python/dazl/ledgerutil/acs.py
```diff
@@ -65,19 +65,49 @@ def widget_json(self, widget_id: str):
...
+        delitem(self._contracts, key)
+
+    def snapshot(self):
+        return ACSView(self, None, dict(self._contracts))
+
+    def strict_snapshot(self):
+        return ACSView(self, self._serial_number, dict(self._contracts))
+
+    def __repr__(self):
+        return f"ACS(len={len(self)}, serial={self._serial_number}, tid={self._template_id})"
```
mschaef-da, line 99: Getting rid of the `empty` constructor made it possible to include the `tid` in the output message, which is probably super useful.

python/dazl/ledgerutil/acs.py
```diff
@@ -107,4 +137,4 @@ def first_payload(self, predicate: ContractMatch) -> Optional[ContractData]:
         return None
 
     def __repr__(self):
-        return f"ACS(len={len(self)})"
+        return f"ACSView(len={len(self)}, snapshot_serial={self._snapshot_serial}, acs={self._acs})"
```
mschaef-da, line 140: Note that the `ACSView`  log message contains both the expected serial number and the ACS (with the current ACS serial number). This makes it possible to log an `ACSView` and see if there's a mismatch that might be causing trouble.

python/dazl/ledgerutil/acs.py
```diff
@@ -65,19 +65,49 @@ def widget_json(self, widget_id: str):
...
+        return f"ACS(len={len(self)}, serial={self._serial_number}, tid={self._template_id})"
+
+
+class ACSView(MutableMapping):
+    def __init__(self, acs: ACS, snapshot_serial: Optional[int], contracts: Dict[ContractId, ContractData]):
+        self._acs = acs
+        self._snapshot_serial = snapshot_serial
+        self._contracts = contracts
+
+    def _maybe_verify_stable_acs(self):
```
da-tanabe, line 108: What calls this? How is this enforced?

mschaef-da:
Part of my motivation for this is work on the `arcade-repo`. I'd like to use an instance of this class as a way to internally maintain a cache of all available arcade artifacts. (It's a small amount of data and potentially used fairly often, so this is a reasonable thing to do.)

What this altered API design does for naive users is two things:

* Mandating the `snapshot` call forces them to at least start understand that there is no such thing as a stable ACS.
* Separating out the `snapshot` into a separate class gives them a Python data structure that's going to be stable for their traversal/etc. work. I've said this in the other PR, but what this means is that errors related to ACS changes will be reported as ledger API errors and NOT Python errors. I strongly believe that the ledger API errors will be much more likely to guide them to an understanding of what's really happening.

For more sophisticated users, this API gives them the choice of how they want to detect ACS stability errors.

python/dazl/ledgerutil/acs.py
mschaef-da: Doh... missed the last step.

python/dazl/ledgerutil/acs.py
```diff
@@ -65,39 +65,76 @@ def widget_json(self, widget_id: str):
...
+        self._maybe_verify_stable_acs()
         return key in self._contracts
 
     def __getitem__(self, key: ContractId) -> ContractData:
+        self._maybe_verify_stable_acs()
         return self._contracts[key]
 
     def __setitem__(self, key: ContractId, payload: ContractData) -> None:
+        self._maybe_verify_stable_acs()
         self._contracts[key] = payload
```
mschaef-da, line 122: This should not be supported at all, really.

python/dazl/ledgerutil/acs.py
```diff
@@ -65,39 +65,76 @@ def widget_json(self, widget_id: str):
...
+        self._maybe_verify_stable_acs()
         return iter(self._contracts)
 
     def __len__(self) -> int:
+        self._maybe_verify_stable_acs()
         return len(self._contracts)
 
     def clear(self) -> None:
+        self._maybe_verify_stable_acs()
         self._contracts.clear()
```
mschaef-da, line 138: Same here... I don't know if Python has `ImmutableMapping`, but this should really be that and not mutable. There is no reason we want to actually support for DAZL users taking a snapshot and then somehow changing it.

python/dazl/ledgerutil/acs.py
```diff
@@ -65,39 +65,76 @@ def widget_json(self, widget_id: str):
...
+    def _set_item(self, key: ContractId, payload: ContractData) -> None:
+        self._serial_number = self._serial_number + 1
+        setitem(self._contracts, key, payload)
+
+    def _del_item(self, key: ContractId) -> None:
+        self._serial_number = self._serial_number + 1
+        delitem(self._contracts, key)
+
+    def snapshot(self):
+        return ACSView(self, None, dict(self._contracts))
```
da-tanabe, line 93: If I'm understanding this correctly, you're proposing that the `ACS` class itself become a bit more opaque, and any read operations must first be done by taking a "snapshot" and then reading off the snapshot.

I'm not sure what I see the utility in `strict_snapshot`; there is simply no safe way to guarantee liveness of the ACS on the client since it's always moving anyway, so if client code is considered about ACS raciness, such error handling _must_ be built regardless:

```
try_again_later = {}
async with ... as conn:
    async for ev in stream.creates():
        try:
            await conn.submit_exercise(ev.contract_id, "SomethingRacy")
        except DazlError:
            try_again_later[ev.contract_id] = ev.payload
```

mschaef-da: There may not be any utility in `strict_snapshot`, the more I think about it. As you point out - It doesn't remove the possibility of the error, it just makes it possible to detect that error earlier (before you do something like issue a ledger command that depends on a now-archived CID).

mschaef-da:
The more I think about it, the more I think this design needs atomic query ops on `ACS` (that might also produce `ACSView` objects with specific subsets of the currently active queries.)

python/dazl/ledgerutil/acs.py
mschaef-da: Handled this a different way - `empty` now takes a TID and the serial number is used to identify connected vs unconnected `ACS` instances.

python/dazl/ledgerutil/acs.py
mschaef-da: No more view.

python/dazl/ledgerutil/acs.py
```diff
@@ -53,58 +54,59 @@ def widget_json(self, widget_id: str):
...
+    def snapshot(self) -> Mapping[ContractId, ContractData]:
+        """
+        Create a snapshot of the _current_ state of the ACS. Useful if you need a consistent view
+        of the ACS at a point in time.
+        """
+        return types.MappingProxyType(dict(self._contracts))
 
     def first_payload(self, predicate: ContractMatch) -> Optional[ContractData]:
-        for value in self._contracts.values():
+        for value in self.snapshot().values():
```
da-tanabe, line 105: I'm not convinced this is necessary…with `asyncio`, the data structure can't be accessed "concurrently" here because there are no `await`'s. Should be fine to have this iterate over `self._contracts.values()` instead.

mschaef-da: How sure are you the predicate won't advance the event loop?

PR #186, opened 2021-03-21
python: Revise the API around template globbing to make it easier to understand

da-tanabe:
python: Revise the API around template globbing to make it easier to understand.

Before this PR, template search/globbing was based on a function called `template_reverse_globs`, which took an evil boolean parameter, and boolean parameters can often make a function's behavior difficult to reason about.

Instead, split this previously ill-defined API into two separate functions:

* `normalize(str | TypeConName) -> str`
  Take a string that might be a template, and convert it to a string of the form `PACKAGE:MODULE:ENTITY`, `PACKAGE:*`, `*:*`, or `*:MODULE:ENTITY`.

* `matching_normalizations(str | TypeConName) -> Sequence[str]`
  Take a string that might be a template filter, and convert it to a list of strings of the form `PACKAGE:MODULE:ENTITY`, `PACKAGE:*`, `*:*`, and `*:MODULE:ENTITY`.


python/dazl/client/events.py
```diff
@@ -62,37 +67,33 @@ def template_reverse_globs(primary_only: bool, package_id: str, type_name: str)
     """
     Return an iterator over strings that glob to a specified type.
     """
-    package_id = package_id or "*"
```
mschaef-da, line 65: I remember seeing this last week and being a touch confused by it. This PR is a nice improvement.

mschaef-da (approved):
Nice.

PR #188, opened 2021-03-23
python: Revision of the ACS API

da-tanabe:
Rewrite the API introduced in #183/#185 to allow for cross-template snapshots.

Usage:

```python
# for "inline" usage
async with ACS(conn, {"SomeMod:SomeTemplate": {}, "OtherTemplate": {}}) as acs:
    # performs a dirty read
    snapshot = acs.read_immediately()
    # these two reads come from the same snapshot so the data is consistent
    cdata1 = snapshot.earliest_contract('SomeMod:SomeTemplate')
    cdata2 = snapshot.earliest_contract('SomeMod:OtherTemplate')

# for "persistent" uasge
class Something:
    def __init__(self, conn):
        self._acs = ACS(conn, {"SomeMod:SomeTemplate": {}})

    async def read_thing(self, widget_id):
        snapshot = await self._acs.read()
        cdata1 = snapshot.earliest_contract('SomeMod:SomeTemplate')
        cdata2 = snapshot.earliest_contract('SomeMod:OtherTemplate')
        return smash_data_together(cdata1, cdata2)
```


mschaef-da:
> ```python
>         self._acs = ACS(conn, {"SomeMod:SomeTemplate": {}})
> 
>     async def read_thing(self, widget_id):
>         snapshot = await self._acs.read()
>         cdata1 = snapshot.earliest_contract('SomeMod:SomeTemplate')
>         cdata2 = snapshot.earliest_contract('SomeMod:OtherTemplate')
> ```

How does `earliest_contract` work for `OtherTemplate` when the `ACS` was only initialized with `SomeTemplate`?

python/dazl/ledgerutil/acs.py
```diff
@@ -2,21 +2,55 @@
...
+from ..query import ContractMatch, is_match
 
-__all__ = ["ACS"]
+__all__ = ["ACS", "Snapshot"]
 
-from ..query import ContractMatch, is_match
 
 T = TypeVar("T")
 
+# Literal types are a bit of a pain to declare, but we _do_ get type safety at the value level
```
mschaef-da, line 45: Yeah... these declarations are a mess. (No offense, I think it's just mypy/python)

python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -517,6 +526,10 @@ async def items(self):
                     else:
                         warnings.warn(f"Received an unknown event: {event}", ProtocolWarning)
                     yield event
+            else:
+                log.debug(
+                    "Not reading from transaction stream because we were only asked for a snapshot."
```
mschaef-da, line 531: This sort of log message makes me super happy.

python/dazl/ledgerutil/acs.py
```diff
@@ -54,59 +88,369 @@ def widget_json(self, widget_id: str):
         class MyService:
             def __init__(self, conn):
                 self._conn = conn
-                self._widgets = ACS.empty("WidgetMod:Widget")
+                self._acs = ACS(conn, {"WidgetMod:Widget": {}})
```
mschaef-da, line 91: Is `ACS.empty` now gone, hopefully?

python/dazl/ledgerutil/acs.py
```diff
@@ -54,59 +88,369 @@ def widget_json(self, widget_id: str):
...
+        """
+        Allow :class:`ACS` to be used within ``async with`` blocks.
+        """
+        await self.stop()
+
+    def __repr__(self):
+        return f"ACS(state={self._state})"
+
+
+async def snapshots(conn: Connection, queries) -> AsyncIterator[Tuple[State, Optional[Snapshot]]]:
```
mschaef-da, line 238: I'd wondered about this sort of thing, and assumed it was a 'v8.1' thing....nice addition.

mschaef-da (approved):
Very nice.

python/dazl/ledgerutil/acs.py
da-tanabe: No it's super gross and I hate it—almost to the point of deciding not to use it. But I like them so much in TypeScript that I keep feeling compelled to give them a chance… :-/

python/dazl/ledgerutil/acs.py
da-tanabe: Yep totally gone—particularly since `ACS` is now more of a stateful object and `Snapshot` takes its place; my original `ACS` class conflated both concerns.

python/dazl/ledgerutil/acs.py
da-tanabe: Yeah it was quite pleasing to wrap the stateful `ACS` class around a simple async iterator that actually handles the "real" work; and its API is pretty enough that arguably, this could _also_ be part of the public API for people that prefer to work with a stream of snapshots rather than a class that "holds" the latest for you.

PR #189, opened 2021-03-24
python: Fix an error in the way that some typing rules are exposed.

da-tanabe:
`typing.Protocol` is actually only available in Python 3.8 and later!


gerolf-da (approved):

gerolf-da:
Thank you so much for this quick fix.

PR #191, opened 2021-03-29
python: Make typing rules for ContractId a little clearer.

da-tanabe:
More typing fixes to other parts of `dazl`.

This was originally an attempt to trace down mypy 0.810 incompatibilities, but the improvements in this PR can be merged in as-is.

~Trying to debug issues where `mypy==0.810` seems considerably stricter than `mypy==0.790`.

For whatever reason recent mypy and current definition of `ContractId` results in an inability to resolve `value` and/or `value_type`, whereas older versions of mypy didn't seem to care.~


python/dazl/cli/metadata.py
```diff
@@ -86,6 +86,9 @@ def _process_metadata(store: PackageStore, options: PrettyOptions):
         formatter = None
 
     pretty_printer = get_pretty_printer(options.format, options, store)
+    if pretty_printer is None:
+        raise RuntimeError(f"unknown format: {options.format}")
```
mschaef-da, line 90: Nice touch, but can this be done within `get_pretty_printer` instead? (Reason being that other pretty printer using CLI commands would also benefit from this error.)

python/dazl/model/types_store.py
```diff
@@ -395,6 +413,10 @@ def __init__(self, mapping: "Mapping[PackageRef, bytes]"):
     def get_package_ids(self) -> "PackageIdSet":
         return frozenset(self.mapping.keys())
 
+    # fetch_package is actually supposed to throw an error on missing packages, but people may
+    # be reliant on the old behavior; anyway this entire file is deprecated; changing the behavior
+    # of a soon-to-be-removed class seems unnecessary
```
mschaef-da, line 418: Agreed.

python/dazl/model/types_store.py
```diff
@@ -428,15 +450,17 @@ class TypeCache(Generic[K, T]):
     """
 
     everything: "Collection[T]"
-    by_package_lookup: "Mapping[str, Mapping[str, Collection[T]]]"
+    by_package_lookup: "Mapping[PackageRef, Mapping[str, Collection[T]]]"
```
mschaef-da, line 453: Nice.

python/dazl/model/writing.py
```diff
@@ -560,7 +560,10 @@ def serialize_command(self, command: "Command") -> "Any":
         elif isinstance(command, ExerciseByKeyCommand):
             name = self.lookup.template_name(command.template_type)
             template = self.lookup.template(name)
-            key_value = self.serialize_value(template.key.type, command.contract_key)
+            key_type = template.key
+            if key_type is None:
+                raise ValueError(f"template {template.tycon} does not have a key")
```
mschaef-da, line 565: Who will know what to do with this error? Seems like there's a lot of distance between this message and whatever needs to be corrected.

python/dazl/util/asyncio_util.py
```diff
@@ -704,7 +723,13 @@ def __repr__(self):
         return f"DeferredStartTask({self._name!r}, state={state})"
 
 
+@no_type_check
 def get_running_loop() -> Optional[AbstractEventLoop]:
+    warnings.warn(
+        "get_running_loop() is deprecated, as support for Python 3.6 will be dropped in dazl 8. "
```
mschaef-da, line 729: Yay!

mschaef-da (approved):

python/dazl/model/writing.py
da-tanabe: It indicates that someone is trying to use `ExerciseByKeyCommand` with a contract that doesn't have a key (so consequently the attempt to serialize the "key" field fails because it doesn't know what to target.

The more problematic part is that this translation happens asynchronously (so the stack trace is fairly useless); this is something that is corrected with the new API.

python/dazl/cli/metadata.py
da-tanabe: I thought about it, but that would require a change to the signature of `get_pretty_printer`. Perhaps something for v8 though; will revisit this on `master`.

PR #195, opened 2021-04-09
python: Typing fixes to some dazl packages

da-tanabe:
Apply fixes as identified by `mypy` to:
* `dazl`
* `dazl.client`
* `dazl.damlast`
* `dazl.prim`
* `dazl.protocols`
* `dazl.server`
* `dazl.values`

And tweak the mypy file to exclude libraries that don't have typing rules.

The bulk of the PR is simply:
* adding type annotations where they were missing
* making the `dazl.damlast` packages better behaved with missing parameters
* mypy isn't wild about functions that have a defined return type but no return statement, so added `return None` in places where it doesn't change behavior, but makes the intent of the method perhaps more obvious.

### DIFF
--- a/python/dazl/client/config.py
+++ b/python/dazl/client/config.py
@@ -5,9 +5,25 @@
 This module contains configuration parameters used by the rest of the library.
 """
 
+# We override types in this file LOTS because mypy does not understand what is happening
+# with the ``config_field`` function. There doesn't appear to be a way to describe
+# a dataclasses-like ``field`` function, which is why this config will be dropped
+# in dazl 8's API.
+
 import argparse
 from dataclasses import asdict, dataclass, field, fields
-from typing import TYPE_CHECKING, Any, Collection, List, Mapping, Optional, Sequence, Type, TypeVar
+from typing import (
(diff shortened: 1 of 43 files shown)


python/dazl/client/config.py
```diff
@@ -148,100 +166,108 @@ class _NetworkConfig(URLConfig):
...
 
-    use_acs_service: bool = config_field(
+    use_acs_service: bool = config_field(  # type: ignore
         "Use Active Contract Set service instead of reading from the Transaction event stream",
         param_type=BOOLEAN_TYPE,
         default_value=True,
     )
 
-    idle_timeout: Optional[float] = config_field(
+    idle_timeout: Optional[float] = config_field(  # type: ignore
```
mschaef-da, line 194: All these `type: ignore`, and this is starting to feel like C.

mschaef-da (approved):

python/dazl/client/config.py
da-tanabe: heh yep… `mypy` basically makes it impossible to build any useful wrappers around `dataclasses`. :-/

PR #197, opened 2021-04-13
python: More typing fixes.

da-tanabe:
Most of these changes are centered around:
* ledger offsets should be typed as `Optional[str]` instead of `str`; we got this wrong in a few places
* internally to `_network_client_impl.py` and `_party_client_impl.py`, some trivial `is not None` checks that would have resulted in much more cryptic `AttributeError: 'NoneType' object has no attribute 'blah'` errors, and now result in slightly less cryptic "you didn't start the client" errors
* the usual bevy of untyped collections or functions that needed type annotations

After this PR, there are only 66 more mypy failures to remedy (down from several hundred prior to the start of this push).


mschaef-da:
> After this PR, there are only 66 more mypy failures to remedy (down from several hundred prior to the start of this push).

Whose woods these are I think I know.   
His house is in the village though;   
He will not see me stopping here   
To watch his woods fill up with snow.   

My little horse must think it queer   
To stop without a farmhouse near   
Between the woods and frozen lake   
The darkest evening of the year.   

He gives his harness bells a shake   
To ask if there is some mistake.   
The only other sound’s the sweep   
Of easy wind and downy flake.   

The woods are lovely, dark and deep,   
But I have promises to keep,   
And miles to go before I sleep,   
And miles to go before I sleep.

    -Robert Frost

python/dazl/client/_party_client_impl.py
```diff
@@ -259,6 +260,9 @@ async def read_acs(
             If the PartyClient is _not_ configured to use the ActiveContractService, this function
             immediately returns `None`, and a resolved future.
         """
+        if self._config is None or self._client_fut is None:
+            raise RuntimeError("read_acs must be called only after a party has started")
```
mschaef-da, line 264: Nice.

mschaef-da (approved):

PR #198, opened 2021-04-13
python: Make mypy happy with every nook and cranny of the dazl codebase.

da-tanabe:
Mostly centered around:

* Invariant enforcement in `ProtobufParser` where `mypy` was correctly flagging some codepaths that would fail with malformed Daml-LF content. (Malformed content will still cause exceptions to be raised, but they'll actually make sense instead of being null pointer exceptions.)
* Deprecating more methods and dropping dependencies on them.
* Removing `dazl.damleval` and `dazl.model.definitions`, as they couldn't really have been used in their current form.
* Backport of #165 to the release branch, as some of the typing fixes were better achieved with some of the new v8 symbols in their new place.

`make typecheck` now runs cleanly on the dazl codebase! Enforcement will be added after the release branch is merged to `master`, and any new code destined for dazl 8 is similarly fixed.


python/dazl/client/pkg_loader.py
```diff
@@ -53,11 +53,11 @@ def __init__(
         self,
         package_lookup: "MultiPackageLookup",
         conn: "SyncPackageService" = None,
-        timeout: "timedelta" = DEFAULT_TIMEOUT,
+        timeout: "Optional[timedelta]" = DEFAULT_TIMEOUT,
     ):
         self._package_lookup = package_lookup
         self._conn = conn
-        self._timeout = timeout
+        self._timeout = timeout if timeout is not None else DEFAULT_TIMEOUT
```
mschaef-da, line 60: Can't this be the following:

```
self._timeout = timeout or DEFAULT_TIMEOUT
```

python/dazl/damlast/pb_parse.py
```diff
@@ -518,7 +528,17 @@ def parse_Scenario_EmbedExpr(self, pb) -> "Scenario.EmbedExpr":
         return Scenario.EmbedExpr(type=self.parse_Type(pb.type), body=self.parse_Expr(pb.body))
 
     def parse_Location(self, pb) -> "Location":
-        return Location(self.parse_ModuleRef(pb.module), self.parse_Location_Range(pb.range))
+        module_ref = self.parse_ModuleRef(pb.module)
+        if module_ref is None:
+            module_ref = self.current_module
+        if module_ref is None:
+            # This is probably a programming mistake, since nowhere in the Daml-LF protobuf does
+            # a Location occur outside of a Module
```
mschaef-da, line 536: Is a `Location` a source code location?

mschaef-da (approved):
LGTM, with one minor comment.

python/dazl/damlast/pb_parse.py
da-tanabe: yup

python/dazl/client/pkg_loader.py
da-tanabe: hah…yeah you're right—will change.

PR #199, opened 2021-04-15
python: Reduce the amount of noise that comes from normal usage of dazl.

da-tanabe:
* Get rid of two warnings that would come about from normal usage.
* Make some of the older dazl `Type` classes more obviously deprecated (PyCharm doesn't seem to mark subclasses as deprecated when their parent classes are deprecated, which is a little annoying.)

### DIFF
--- a/python/dazl/model/types.py
+++ b/python/dazl/model/types.py
@@ -32,6 +32,7 @@ from typing import (
     AbstractSet,
     Any,
     Callable,
+    ClassVar,
     Collection,
     Dict,
     Mapping,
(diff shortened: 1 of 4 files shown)


python/dazl/model/types.py
```diff
@@ -82,27 +83,35 @@ def type_ref(s: str) -> "TypeReference":
...
-        components = s.split(":")
-        if len(components) != 3:
-            raise ValueError(f"could not parse as a template reference: {s!r}")
+    warnings.warn(
+        "type_ref is deprecated; there is no replacement", DeprecationWarning, stacklevel=2
+    )
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", DeprecationWarning)
+
+        # Attempt to parse in an older format still used by Navigator for display purposes.
```
mschaef-da, line 92: We need to get Navigator to change to the new format...

mschaef-da (approved):
LGTM

python/dazl/model/types.py
da-tanabe: heh yeah agreed…

PR #200, opened 2021-04-15
python: Add a test case for explicit package loading.

da-tanabe:


da-tanabe:
This was incorporated in #201.

PR #204, opened 2021-04-19
python: Add more typing rules.

da-tanabe:
In trying to backport the new API from `master`  to the v7 branch, I discovered there were still more bugs in the typing rules for Protobuf-generated code:
* `command_service.proto`'s rules were missing altogether; I had skipped over it thinking that I wasn't using it, but that was only true of the old API. The new API uses it (which is part of the point of the new API!), so add stubs for these too.
* The typing rules pulled in the wrong `Party` class; the `Party` literal was meant to reference `NewType(str, 'Party')`, not the AST type that defines a `Party` in Daml-LF.
* `Event` was missing a handful of fields; added those in.
* The return type for stream calls is a little…complicated. And since we actually use `Call.cancel()` in our code the return types had to be typed appropriately too.
* The `grpc-stubs` library's `grpc.aio` support isn't _quite_ ready for prime time, so `grpc.aio` has to be excluded from mypy checks. Thankfully, the crazy games we play with constructor overloading still do the right thing and automatically make stubs `async` when given a `grpc.aio.Channel`, and non-async when given a `grpc.Channel`.


python/dazl/_gen/com/daml/ledger/api/v1/event_pb2.pyi
```diff
@@ -16,8 +16,9 @@ from typing import (
 
 from google.protobuf.descriptor import FieldDescriptor as __FieldDescriptor
 from google.protobuf.message import Message as __Message
+from google.protobuf.wrappers_pb2 import StringValue
 
-from .......damlast.daml_types import Party
+from .......prim import Party
```
mschaef-da, line 21: I've heard of 'three star programmers' before, but never a _seven dot_ programmer....

https://wiki.c2.com/?ThreeStarProgrammer

da-tanabe: hahahaha yeah these aren't awesome…

I don't know what to do with these very very primitive types, but I definitely hate these imports too. Know that it is something on my mind… :-)

PR #220, opened 2021-04-23
samples: Update Pipfile.lock files.

da-tanabe:
The lockfiles point to some fairly outdated dependencies, so simply run `pipenv update`.

The codebase has pretty much moved on to poetry, but I'd rather silence the security warnings now rather than blocking this on converting samples over to poetry.


da-tanabe:
Actually thinking about this more, there's no way I can update the Pipfile any more, because it doesn't know how to read dependencies from the poetry file. Closing…

PR #225, opened 2021-04-27
python: Backport typing fixes for dazl.server from the v7 branch.

da-tanabe:
A few type hints taken from the v7 branch and backported to `master`.

### DIFF
--- a/python/dazl/server/management.py
+++ b/python/dazl/server/management.py
@@ -6,8 +6,7 @@ Endpoints for managing parties/bots connected via a dazl client.
 """
 
 from dataclasses import asdict, dataclass
-from typing import TYPE_CHECKING, Collection, Sequence
-import warnings
+from typing import TYPE_CHECKING, Collection, NoReturn, Optional, Sequence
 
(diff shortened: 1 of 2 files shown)


python/dazl/server/management.py
```diff
@@ -70,7 +67,7 @@ async def get_bot_by_id(request: "web.Request") -> "web.Response":
         return web.json_response(asdict(_get_bot_info(bot)))
 
     @routes.post("/bots/{bot_id}/state/{action}")
-    async def change_bot_state(request: "web.Request") -> "web.Response":
+    async def change_bot_state(request: "web.Request") -> "NoReturn":
```
mschaef-da, line 70: Why is this raising the response and not returning it?

da-tanabe: Hmm that's a good question…I don't actually know. Will look into this further…

(Certainly the existing change is all that was needed to make mypy happy, but you're right—this code _does_ look a little weird.)

mschaef-da (approved):

PR #242, opened 2021-05-07
python: Fix bugs in client-side filtering on the new API

da-tanabe:
python: In trying to document `dazl.ledger.Connection`'s `query`, `stream`, `query_many`, and `stream_many` methods, I realized the API was a bit underspecified, and additionally could also be made to be slightly less clunky; and then as I dug in even deeper, I realized that there were bugs in the implementation

* Fix client-side filtering for the new streaming methods.
* Make dazl.query's API a little clearer in preparation for better documentation.
* Wrap some tests around it.


python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -495,20 +496,22 @@ def query(self, __template_id: str = "*", __query: Query = None) -> "QueryStream
         :param __query:
             A filter to apply to the set of returned contracts.
         """
-        return QueryStream(self, {__template_id: __query}, UNTIL_END)
+        return QueryStream(
+            self, parse_query({__template_id: __query}, server_side_filters=False), UNTIL_END
+        )
 
-    def query_many(self, __queries: Optional[Mapping[str, Query]] = None) -> "QueryStream":
+    def query_many(self, *queries: Queries) -> "QueryStream":
```
mschaef-da, line 503: I like this signature a lot better.

python/dazl/query/__init__.py
```diff
@@ -1,6 +1,135 @@
...
+
+ContractMatch = Union[None, Callable[[ContractData], bool], ContractData]
+
+TemplateName = Union[str, "TypeConName"]
+Query = Union[None, ContractData, Callable[[ContractData], bool]]
+Queries = Union[None, TemplateName, Collection[TemplateName], Mapping[TemplateName, Query]]
+
+
+class Filter:
+    server_side: Optional[ContractData]
```
mschaef-da, line 21: Is there ever a filter that has both a server and a client side?  Based on this, I think no?:

https://github.com/digital-asset/dazl-client/pull/242/files#diff-02769ebb8d208465ff7c62ac5560aa3442e70f58cde707fd6b60da6c2adf9e95R90

Given how distinct a client side filter is from a server side filter, I'd really like to see these as two distinct types with printable representations that indicate which is which. From DAZL's PoV, these two types of filters may originate from the same place in the API, and have similar effects to the contents of the output stream, but they're really totally different concepts.

mschaef-da (approved):

python/dazl/ledger/grpc/conn_aio.py
da-tanabe: yep…this is exactly where I fell into the rabbit hole… :-)

python/dazl/query/__init__.py
da-tanabe: This is somewhat in anticipation of the HTTP JSON API…it may very well change as I get into the weeds of it.

Issue #250, opened 2021-05-19
`dazl.ledgerutil.ACS` should not attempt to retry reading from a closed connection


da-tanabe:
Closed connections confuse the retry logic in `dazl.ledgerutil.ACS`; these kinds of errors should never be retried, because once a channel is closed, it can never be used again.

```
2021-05-19T20:11:17+0000 [ERROR] (dazl.conn.5873) The stream disconnected. Waiting 240.0 seconds, then trying again (attempting to resume at offset 000000000058d372)
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/dazl/ledgerutil/acs.py", line 252, in snapshots
    async for obj in stream:
  File "/usr/local/lib/python3.8/site-packages/dazl/ledger/grpc/conn_aio.py", line 700, in items
    async for event in self._tx_events(tx_filter_pb, offset):
  File "/usr/local/lib/python3.8/site-packages/dazl/ledger/grpc/conn_aio.py", line 748, in _tx_events
    self._response_stream = response_stream = stub.GetTransactions(request)
  File "/usr/local/lib/python3.8/site-packages/grpc/aio/_channel.py", line 154, in __call__
    call = UnaryStreamCall(request, deadline, metadata, credentials,
  File "/usr/local/lib/python3.8/site-packages/grpc/aio/_call.py", line 532, in __init__
    channel.call(method, deadline, credentials, wait_for_ready),
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/channel.pyx.pxi", line 123, in grpc._cython.cygrpc.AioChannel.call
grpc._cython.cygrpc.UsageError: Channel is closed.
```

PR #251, opened 2021-05-20
python: Address a problem that could cause ACS views to incorrectly persistently retry

da-tanabe:
Address a problem that could cause ACS views to persistently retry, even when the underlying connection was closed (addresses #250).


python/dazl/ledgerutil/acs.py
```diff
@@ -289,7 +293,7 @@ async def snapshots(
                 offset,
             )
             await sleep(backoff_time)
-            backoff_time = max(30.0, backoff_time * 2.0)
+            backoff_time = min(30.0, backoff_time * 2.0)
```
mschaef-da, line 296: Ick. Good catch.

mschaef-da (approved):

python/dazl/ledgerutil/acs.py
da-tanabe: hah…yep—I was wondering how I was getting 240 second backoffs… :-P

PR #253, opened 2021-05-24
python: Add some test utilities.

da-tanabe:
In trying to fix another bug, I realized how verbose the code around writing tests was, and set out to try to reduce it.

The total amount of code is higher than I would have liked, but that's mostly because of Python 3.6 support; it could have been a lot less if I could have used `contextlib.asynccontextmanager` and `contextlib.AsyncExitStack`, but unfortunately those were introduced in Python 3.7.

### DIFF
--- a/python/dazl/testing/connect.py
+++ b/python/dazl/testing/connect.py
@@ -7,6 +7,7 @@ import sys
 from typing import BinaryIO, Callable, Collection, List, Optional, Sequence, Union
 
 from .. import connect
+from ..ledger import PartyInfo
 from ..ledger.aio import Connection
 from ..prim import Party
 
@@ -168,22 +169,7 @@ class ConnectWithNewParty:
         self.connections = []  # type: List[Connection]
 
     async def __aenter__(self) -> "Sequence[ConnectionWithParty]":
-        # Party allocation and package uploading require "admin" access, so we disregard the given
-        # user parameters to allocate parties
-        async with connect(url=self.url, admin=True, ledger_id=self.ledger_id) as conn:
-            party_infos = await gather(
-                *(
-                    conn.allocate_party(
-                        identifier_hint=self.identifier_hint_fn(i),
-                        display_name=self.display_name_fn(i),
-                    )
-                    for i in range(self.party_count)
-                )
-            )
-
-            if self.dar is not None:
-                await conn.upload_package(self.dar)
-
+        party_infos = await self._set_up_ledger()
         ret = []
         for party_info in party_infos:
             party_conn = connect(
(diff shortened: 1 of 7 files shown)


python/dazl/testing/connect.py
```diff
@@ -0,0 +1,204 @@
...
+                    conn.allocate_party(
+                        identifier_hint=self.identifier_hint_fn(i),
+                        display_name=self.display_name_fn(i),
+                    )
+                    for i in range(self.party_count)
+                )
+            )
+
+            if self.dar is not None:
+                await conn.upload_package(self.dar)
```
mschaef-da, line 185: Given that this `async with` happens inside the beginning of an enclosing `async with`, I'd be sorely tempted to make this its own function.

python/dazl/testing/connect.py
```diff
@@ -0,0 +1,204 @@
...
+                url=self.url,
+                read_as=self.read_as,
+                act_as=[party_info.party, *self.act_as],
+                admin=self.admin,
+                ledger_id=self.ledger_id,
+            )
+            await party_conn.open()
+            ret.append(ConnectionWithParty(party_conn, party_info.party))
+        if len(ret) == 1:
+            return ret[0]
```
mschaef-da, line 199: Is this right? The function is declared to return a `Sequence[...]`, and this is returning a `ConnectionWIthParty`.

da-tanabe: This is a(n admittedly annoying) style I picked up from numpy for functions that expect to be splatted on return.

```py
a = return_values(count=1)
a, b = return_values(count=2)
a, b, c = return_values(count=3)
```

In the first case, the signature is `T`; in _every other case_, the signature is a tuple of the required number of elements:

```py
>>> reveal_type(return_values(count=1))
T
>>> reveal_type(return_values(count=2))
Tuple[T, T]
>>> reveal_type(return_values(count=3))
Tuple[T, T, T]
```

It makes writing the code in a generic way very obnoxious; however, it makes the call site much more pleasant to work with: 

Single party:

https://github.com/digital-asset/dazl-client/blob/65315ada2948803088b081373872f1afe8700d38/python/tests/unit/test_ledgerutil_acs_close.py#L12

Multiple party:

https://github.com/digital-asset/dazl-client/blob/65315ada2948803088b081373872f1afe8700d38/python/tests/unit/test_protocol_ledgerapi.py#L15

If the signature were really constrained to requiring only a collection, then in the first case, you could _only_ write:

```py
async with connect_with_new_party(url=sandbox, dar=PostOffice) as (p,):
```

There are some pretty significant typing rules in the companion `connect.pyi` file that declare the "intended" types.

python/dazl/testing/connect.py
da-tanabe: Yep on second look I agree with you; this is sufficiently "different" from the rest of the function that it warrants its own thing. I might also convert the second half just for good measure too.

mschaef-da (approved):

PR #262, opened 2021-07-07
python: Add support for Daml-LF 1.14 archives.

da-tanabe:
python: Add support for Daml-LF 1.14 archives.

There are some minor under-the-hood changes too:

* `dazl/_gen/google/rpc/status_pb2.py` was never actually used, so it has been removed in favor of Google's code. #260
* Now that we don't need to consider a mixture of raw `.proto` files and `.zip` files, `_build/unpack.py` was made a bit simpler, with an eye towards eventually replacing the entire script with a simple `unzip` call.
* Instead of using Daml-LF Dev, dazl now uses Daml-LF 1.14 specifically; subsequent changes in `dazl` where we upgrade Daml-LF will stick with using the most recent non-dev version.

The Protobuf/gRPC code is in bad need of an overhaul to further simplify things, but here I wanted to limit scope to the bare minimum required to support Daml-LF 1.14 archives.


_build/unpack.py
```diff
@@ -48,83 +49,60 @@ def run(self) -> 'Manifest':
...
+                raise ValueError(f"don't know how to process {input_path}")
         
         return manifest
     
     def _process_zip(self, path: 'Path') -> 'Manifest':
-        proto_packages = {}  # type: Manifest
+        proto_packages = {}  # type: Dict[str, Sequence[str]]
         with ZipFile(path) as z:
-            logging.info('Procesing zip file: %s', path)
+            logging.info('Processing zip file: %s', path)
```
mschaef-da, line 60: 👍

python/dazl/damlast/pb_parse.py
```diff
@@ -285,6 +290,8 @@ def parse_Expr(self, pb) -> "Expr":
             args["to_text_template_id"] = self.parse_Expr_ToTextTemplateId(pb.to_text_template_id)
         elif sum_name == "type_rep":
             args["type_rep"] = self.parse_Type(pb.type_rep)
+        elif sum_name == "throw":
+            args["throw"] = self.parse_Expr_Throw(pb.throw)
```
mschaef-da, line 294: Glad to see this...

mschaef-da (approved):

PR #265, opened 2021-07-29
python: Fix an error in the way that dazl interprets JWTs.

da-tanabe:
python: Fix an error in the way that dazl interprets JWTs; it was incorrectly assuming that Python would handle unpadded base64 strings in a JWT.

Separately address a problem where the gRPC `authorization` header was being sent with a capital letter, which although valid HTTP/1.1, is apparently not valid gRPC.


nycnewman (approved):
LGTM

PR #278, opened 2021-08-21
python: Fixed some leaky deprecation warnings

da-tanabe:
Fixed some leaky deprecation warnings that would happen even when deprecated types weren't explicitly imported.

### DIFF
--- a/python/dazl/model/__init__.py
+++ b/python/dazl/model/__init__.py
@@ -21,7 +21,9 @@ introduced in dazl v5) or :mod:`dazl.protocols` (for the API introduced in dazl 
 
 import warnings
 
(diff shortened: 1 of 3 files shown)


python/dazl/model/__init__.py
```diff
@@ -21,7 +21,9 @@
 
 import warnings
 
-from . import core, ledger, lookup, network, reading, types, writing
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore", DeprecationWarning)
+    from . import core, ledger, lookup, network, reading, types, writing
```
mschaef-da, line 26: Dunno why, but this `import` line strikes me as a particularly pleasing line of code

mschaef-da (approved):

PR #282, opened 2021-09-11
python: Add support for complex types (records, variants, or other maps) as the keys to a GenMap.

da-tanabe:
python: Add support for complex types (records, variants, and another genmaps) as the keys for a GenMap.


leonidr-da (approved):
Excellent.

PR #283, opened 2021-09-19
python: Typing files fixes.

da-tanabe:
python: Fixes to typing files.

dazl's mypy.ini file formerly _excluded_ `dazl._gen` for historical reasons, but since typings files were introduced for these files in #192 and #271, this exclusion should have been removed. But as it turns out, these exclusions were hiding the fact that there were some bugs in the typing rules, notably around Python 3.6/3.7 support. This PR removes those exclusions and fixes bugs in the typing files.


python/dazl/protocols/v1/pb_parse_event.py
```diff
@@ -239,18 +239,14 @@ def serialize_transactions_request(
...
+        ledger_offset = lapipb.LedgerOffset(boundary=0)
 
     if f.destination_offset is not None:
-        final_offset = lapipb.LedgerOffset()
-        final_offset.absolute = f.destination_offset
+        final_offset = lapipb.LedgerOffset(absolute=f.destination_offset)
     else:
-        final_offset = lapipb.LedgerOffset()
-        final_offset.boundary = 1
+        final_offset = lapipb.LedgerOffset(boundary=1)
```
mschaef-da, line 249: Typing issues aside, I like this better anyway.

python/dazl/protocols/v1/pb_parse_event.py
```diff
@@ -432,11 +428,12 @@ def to_event(
     context: "Union[TransactionEventDeserializationContext, ActiveContractSetEventDeserializationContext]",
     evt_pb: "Union[lapipb.Event, lapipb.TreeEvent]",
 ) -> "Optional[OffsetEvent]":
+    # TODO: mypy should understand that
     try:
-        event_type = evt_pb.WhichOneof("event")
+        event_type = evt_pb.WhichOneof("event")  # type: ignore
```
mschaef-da, line 433: Not so excited about this.

mschaef-da (approved):
LGTM

python/dazl/protocols/v1/pb_parse_event.py
da-tanabe: indeed

python/dazl/protocols/v1/pb_parse_event.py
da-tanabe: yeah I don't know why `mypy` has trouble with the `Union` in this context. Nevertheless, this whole file is pending deprecation anyway thank goodness.

PR #306, opened 2021-12-04
meta: Upgrade all fixtures and samples to Daml Connect 1.17.1.

da-tanabe:
meta: Upgrade all fixtures and samples to Daml Connect 1.17.1.

### DIFF
--- a/samples/ping-pong/daml.yaml
+++ b/samples/ping-pong/daml.yaml
@@ -1,4 +1,4 @@
-sdk-version: 0.13.12
+sdk-version: 1.17.1
 name: pingpong
 source: daml/PingPong.daml
 parties:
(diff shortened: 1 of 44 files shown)


samples/ping-pong/daml.yaml
```diff
@@ -1,4 +1,4 @@
-sdk-version: 0.13.12
+sdk-version: 1.17.1
```
mschaef-da, line 1: This must be something of a record for gap between SDK versions...

mschaef-da (approved):

samples/ping-pong/daml.yaml
da-tanabe: Thank goodness I had no models to upgrade… :-P

PR #312, opened 2022-01-03
meta: Update copyright notices for the year 2022.

da-tanabe:
Update copyright notices for the year 2022.


mschaef-da (approved):
Nice.. 
image

PR #330, opened 2022-03-07
python: Introduce testing utilities for authorization and TLS.

da-tanabe:
python: Introduce testing utilities for authorization and TLS.

* Bump up OpenSSL and cryptography dependencies—they were fairly ancient.
* Add a test that ensures that TLS support works.
* Minor changes to token handling to avoid sending locally-manufactured tokens.


mschaef-da (approved):
LGTM.

PR #331, opened 2022-03-07
meta: Correct a configuration error that prevented Daml 2.0 tests from actually running.

da-tanabe:
meta: Correct a configuration error that prevented Daml 2.0 tests from actually running.

Also addressed:
* rewrote the Daml models to use Daml 2.0 syntax

#### Notes

In #329, I accidentally used `env` instead of `environment` in the CircleCI config file, which meant the dual-SDK-version tests weren't running as expected.

After fixing the initial error, I played around with a few different ways of triggering the tests (including tests that are, by design, supposed to only work with one or the other version). I stumbled across parameterized fixtures, which also solves nicely for being able to run the tests locally, so I switched to that instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,8 +58,6 @@ jobs:
     parameters:
       python_version:
         type: string
-      daml_sdk_version:
-        type: string
     docker:
       - image: cimg/python:<<parameters.python_version>>-browsers
     resource_class: medium+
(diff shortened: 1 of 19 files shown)


jordan-da (approved):
build stuff looks ok to me

alexgraham-da (approved):
looks good to me 👍

PR #335, opened 2022-03-31
python: Minor fixes to user-related APIs.

da-tanabe:
python: Fix a bug where the `Connection.create_user` method did not send a `CreateUserRequest` properly if no rights were specified.

Also some minor changes to the test structure to allow specific tests to be V2-only, as the user API is only available in V2 and later.


python/dazl/ledger/grpc/conn_aio.py
```diff
@@ -709,13 +709,10 @@ async def get_user(self, user_id: "Optional[str]" = None) -> "User":
...
-        request = (
-            lapiadminpb.CreateUserRequest(
-                user=Codec.encode_user(user), rights=[Codec.encode_right(right) for right in rights]
-            )
-            if rights
-            else ()
-        )
+        request = lapiadminpb.CreateUserRequest(user=Codec.encode_user(user))
+        if rights is not None:
+            request.rights.extend(Codec.encode_right(right) for right in rights)
```
mschaef-da, line 714: It feels a little odd to see an encoding being passed into `extend`... seems like the encoding should more of a concern at the root of the data structure.

python/tests/conftest.py
```diff
@@ -25,7 +25,21 @@ def sandbox(request) -> "Generator[str, None, None]":
...
+
+
+@pytest.fixture(scope="session")
+def sandbox_1() -> "Generator[str, None, None]":
+    with testing.sandbox(project_root=None, version="1.18.1") as sb:
+        yield sb.url
+
+
+@pytest.fixture(scope="session")
+def sandbox_2() -> "Generator[str, None, None]":
```
mschaef-da, line 41: Might be nice to have these `_v1` and `_v2`, otherwise it might as well be this (two equivalent but distinct entities):

image

mschaef-da (approved):

python/dazl/ledger/grpc/conn_aio.py
da-tanabe: It's really just because I didn't bother to create a POPO (plain-ol Python object) for `CreateUserRequest`, so it didn't make sense to have a `Codec.encode_create_user_request` function either. I may decide to do that one day if this function bloats further, but for now I don't see that as too likely.

python/tests/conftest.py
da-tanabe: ^ that's not exactly dissuading me… :-P

PR #338, opened 2022-05-16
python: Add support for VersionService.

da-tanabe:
python: Add support for `VersionService` to allow gRPC Ledger API clients to inspect feature availability.


python/tests/unit/test_ledger_api_version.py
```diff
@@ -0,0 +1,18 @@
...
+
+@pytest.mark.asyncio
+async def test_ledger_api_version(sandbox):
+    async with connect(url=sandbox, admin=True) as conn:
+        version = await conn.get_version()
+
+        logging.info("Ledger API version: %s", version.version)
+        logging.info(
+            "User management API supported: %s", version.features.user_management.supported
+        )
```
sammy-da, line 18: logs?

da-tanabe: Yep, I have the standard Python logger wired into pytest so this stuff ends up with the associated test!

sammy-da (approved):

PR #332, opened 2022-03-10
python: Add support for fetching rights from a Daml 2.0 ledger.

da-tanabe:
python: Add support for fetching rights from a Daml 2.0 ledger.

This also includes minor fixes for running the test suite against a Daml 2.0 ledger.

PR #340, opened 2022-05-17
python: Fix errors in processing Daml V1 tokens when submitting commands.

da-tanabe:
python: Fix a problem where submitting commands with Daml V1 tokens were being processed incorrectly.

Fixes a regression introduced by #332.

The majority of this PR is actually about introducing some utilities for pretty-printing for `__repr__` functions, as careful inspection of better `__repr__` definitions for some key objects helped to narrow down where party information from tokens was getting dropped on the floor.


python/dazl/ledger/api_types.py
```diff
@@ -407,6 +409,18 @@ def __eq__(self, other: Any) -> bool:
...
+    def __repr__(self) -> str:
+        s = []
+        if self.workflow_id is not None:
+            s.append(f"workflow_id={_repr.str(self.workflow_id)}")
+        if self.command_id is not None:
+            s.append(f"command_id={_repr.str(self.command_id )}")
+        if self.read_as is not None:
+            s.append(f"read_as={_repr.list(self.read_as)}")
+        if self.act_as is not None:
+            s.append(f"act_as={_repr.list(self.act_as)}")
```
mschaef-da, line 421: I wonder if these should be explicitly displayed with some sort of 'none' signifier if they're missing.

python/dazl/ledger/config/access.py
```diff
@@ -291,6 +295,21 @@ def application_name(self) -> str:
     def token_version(self) -> "Literal[1, 2]":
         return self._token_version
 
+    def __repr__(self):
+        s = [
+            f"read_as={_repr.list(self.read_as)}",
+            f"act_as={_repr.list(self.act_as)}",
+        ]
+        if self.admin:
+            s.append("admin=True")
```
mschaef-da, line 304: Same thing here.... why is `admin=False` not displayed?

mschaef-da (approved):

python/dazl/ledger/api_types.py
da-tanabe: I actually intentionally left them out just to keep the `repr` short; more often than not they're _not_ defined so logging will be a little less noisy being selective about what's logged.

PR #344, opened 2022-05-31
meta: Upgrade Protobuf, and regenerate code for compatibility with newer versions of Protobuf.

da-tanabe:
meta: Regenerate Protobuf files for compatibility with newer versions of protobuf libraries.


go/api/com/daml/ledger/api/v1/command_completion_service_grpc.pb.go
```diff
@@ -3,7 +3,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.2.0
-// - protoc             v3.18.1
+// - protoc             v3.19.4
```
dasormeter, line 6: daml repo is 3.19.3 -- probably not a biggie but why not match up with it?

go/api/com/daml/ledger/api/v1/event.pb.go
```diff
@@ -389,6 +389,7 @@ type ExercisedEvent struct {
 	WitnessParties []string `protobuf:"bytes,10,rep,name=witness_parties,json=witnessParties,proto3" json:"witness_parties,omitempty"`
 	// References to further events in the same transaction that appeared as a result of this ``ExercisedEvent``.
 	// It contains only the immediate children of this event, not all members of the subtree rooted at this node.
+	// The order of the children is the same as the event order in the transaction.
```
dasormeter, line 392: Intentional added comment?

da-tanabe: These get generated directly from the Protobuf files, so as the `.proto` files from the daml repo change, these will change as well (essentially whenever codegen is rerun, like it is in this PR).

go/api/com/daml/ledger/api/v1/command_completion_service_grpc.pb.go
da-tanabe: The version of `protoc` here is pulled in transitively through `grpcio-tools`, so it's somewhat non-trivial to force a specific version here; and since dazl is a library, it would be unfriendly to force-pin a specific version as that would then transitively apply to all libraries that depend on dazl.

poetry.lock
```diff
@@ -1049,7 +1049,7 @@ tls-testing = ["pyOpenSSL"]
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.6.2"
-content-hash = "4154da8020188f4e7e6575914d20096acb8f9f6ade61aa9dd754805d3661fd94"
+content-hash = "66288dd23101e028a2f81c95c56093aff80a22502f6820b6c88a585ab46d44c7"
```
dasormeter, line 1052: hmm why did hash change when version did not?

python/dazl/_gen/com/daml/daml_lf_1_14/daml_lf_1_pb2.py
```diff
@@ -18,777 +19,13 @@
...
...x12\x11\n\rEQUAL_NUMERIC\x10x\x12\x0e\n\nEQUAL_TEXT\x10Q\x12\x13\n\x0f\x45QUAL_TIMESTAMP\x10R\x12\x0e\n\nEQUAL_DATE\x10S\x12\x0f\n\x0b\x45QUAL_PARTY\x10T\x12\x0e\n\nEQUAL_BOOL\x10U\x12\x15\n\x11\x45QUAL_CONTRACT_ID\x10V\x12\x0e\n\nEQUAL_LIST\x10W\x12\x12\n\x0e\x45QUAL_TYPE_REP\x10{\x12\n\n\x05\x45QUAL\x10\x83\x01\x12\x0c\n\x07LESS_EQ\x10\x84\x01\x12\t\n\x04LESS\x10\x85\x01\x12\x0f\n\nGREATER_EQ\x10\x86\x01\x12\x0c\n\x07GREATER\x10\x87\x01\x12\t\n\x05TRACE\x10X\x12\x16\n\x12\x43OERCE_CONTRACT_ID\x10\x66\x12\x17\n\x13\x43ODE_POINTS_TO_TEXT\x10i\x12\x17\n\x13TEXT_POINTS_TO_CODE\x10j\x12\x15\n\x10SCALE_BIGNUMERIC\x10\x89\x01\x12\x19\n\x14PRECISION_BIGNUMERIC\x10\x8a\x01\x12\x13\n\x0e\x41\x44\x44_BIGNUMERIC\x10\x8b\x01\x12\x13\n\x0eSUB_BIGNUMERIC\x10\x8c\x01\x12\x13\n\x0eMUL_BIGNUMERIC\x10\x8d\x01\x12\x13\n\x0e\x44IV_BIGNUMERIC\x10\x8e\x01\x12\x1b\n\x16SHIFT_RIGHT_BIGNUMERIC\x10\x8f\x01\x12\x1a\n\x15\x42IGNUMERIC_TO_NUMERIC\x10\x90\x01\x12\x1a\n\x15NUMERIC_TO_BIGNUMERIC\x10\x91\x01\x12\x17\n\x12\x42IGNUMERIC_TO_TEXT\x10\x92\x01\x42}\n\x15\x63om.daml.daml_lf_1_14ZDgithub.com/digital-asset/dazl-client/v7/go/api/com/daml/daml_lf_1_14\xaa\x02\x1d\x43om.Daml.Daml_Lf_1_14.DamlLf1b\x06proto3')
```
dasormeter, line 22: why such a significant structural change here?

dasormeter (approved):
many files changed.  much code generation.

poetry.lock
da-tanabe: I specified a new lower bound for protobuf libraries (dazl now requires at least Protobuf 3.19.x), so that changes the hash of `pyproject.toml`.

PR #347, opened 2022-07-12
python: Add Connection.get_ledger_end.

da-tanabe:
python: Add a `Connection.get_ledger_end` call for retrieving the current ending offset for a ledger.


alexgraham-da (approved):
Nice, lgtm

PR #348, opened 2022-07-18
python: Have every file treat type annotations properly when there are forward references to types.

da-tanabe:
python: Have every file treat type annotations properly when there are forward references to types.

In particular, add `from __future__ import annotations`, which delays type annotation evaluation.


alexgraham-da (approved):
👍

PR #358, opened 2022-09-28
python: Avoid using pkg_resources.

da-tanabe:
`dazl` currently relies on `setuptools` to be installed. This is an exceedingly common situation, but with Poetry 1.2 no longer implicitly doing anything with `setuptools`, may be less common in the future. Remove the call to `pkg_resources` because it's actually not part of the _standard_ Python library, and is overkill to avoid hardcoding One More Version String™.


alexgraham-da (approved):
LGTM, the test is a nice touch 👍

PR #346, opened 2022-06-28
add an upper limit to protobuf

alexgraham-da:
As of version `7.8.0`, bots that were automatically installing version `4.21.2` of `protobuf` would crash with:

```
[   INFO] 2022-06-28 14:59:34,136 | dazl    | Parsed 295289 bytes of metadata (package ID '108aa371793a39379d09dd6f02b5613e8bf37ec9893a826747545df5318475f1') in 1.63 ms.
[   INFO] 2022-06-28 14:59:34,139 | dazl    | Parsed 165471 bytes of metadata (package ID '0323a524706f5ab0c24f300e46879f70332debfe5cc405aac1fe958285bc35fa') in 0.84 ms.
[  ERROR] 2022-06-28 14:59:34,148 | dazl    | The main monitoring thread died.
Traceback (most recent call last):
  File "/Users/alexgraham/.local/share/virtualenvs/exberry_adapter-tsmRtwmt/lib/python3.7/site-packages/dazl/protocols/autodetect.py", line 182, in _main
    for i, metadata in enumerate(_monitor_ledger_network(conn)):
  File "/Users/alexgraham/.local/share/virtualenvs/exberry_adapter-tsmRtwmt/lib/python3.7/site-packages/dazl/protocols/v1/grpc.py", line 270, in grpc_main_thread
    grpc_package_sync(package_provider, store)
  File "/Users/alexgraham/.local/share/virtualenvs/exberry_adapter-tsmRtwmt/lib/python3.7/site-packages/dazl/protocols/v1/grpc.py", line 365, in grpc_package_sync
    metadatas_pb[package_id] = parse_archive_payload(package_id, archive_payload)
  File "/Users/alexgraham/.local/share/virtualenvs/exberry_adapter-tsmRtwmt/lib/python3.7/site-packages/dazl/damlast/parse.py", line 58, in parse_archive_payload
    archive_payload.ParseFromString(archive_bytes)
google.protobuf.message.DecodeError: Error parsing message
[  ERROR] 2022-06-28 14:59:34,148 | dazl    | The above error was propagated as an initialization error.
[   INFO] 2022-06-28 14:59:34,148 | dazl    | Closing a GRPCv1Connection channel...
Traceback (most recent call last):
  File "bot/exberry_adapter_bot.py", line 558, in <module>
    main()
  File "bot/exberry_adapter_bot.py", line 538, in main
    network.run_forever()
[...]

PR #359, opened 2022-09-28
python: Make DAR parsing compatible with bigger DALFs.

da-tanabe:
python: Make DAR parsing compatible with bigger DALFs.

By default, the new Protobuf parsing library `upb` that sits underneath `protobuf>=4` will refuse to parse messages that are deeper than 100 ([protocolbuffers/upb:python/message.c](https://github.com/protocolbuffers/upb/blob/edecfd5eb06ee7a55206d4651222d03b0a4f7b01/python/message.c#L1239)), and unfortunately the vast majority of Daml models will end up with messages that easily exceed that depth. Use the undocumented `SetAllowOversizeProtos` method (which was carried over from the older C++ implementation) in order to allow these "larger" Protobuf objects to be parsed.

This is ultimately the fix to the issue that #346 avoided simply by stating dazl was only compatible with Protobuf 3.


alexgraham-da (approved):
👍 fun

PR #362, opened 2022-10-18
python: Add a timeout parameter to `Connection` method calls

da-tanabe:
python: Add a timeout parameter to method calls to allow for specifying per-call timeouts, and for establishing Connection-scoped timeouts.


python/dazl/ledger/_retry.py
```diff
@@ -33,8 +40,10 @@ async def retry(
     while True:
         try:
             return await fn()
+        except CancelledError:
+            raise
         except Exception as ex:
-            if (datetime.now() - start) <= timeout and retry_safe_ex(ex):
+            if (datetime.utcnow() - start) <= timeout and retry_safe_ex(ex):
```
mschaef-da, line 46: Why the switch here?

mschaef-da (approved):

python/dazl/ledger/_retry.py
da-tanabe: This actually fixes a horrible bug… a few lines above what GH is showing in the diff, you can see that `start` is originally set to `datetime.utcnow()`, so I was comparing apples with… offset apples incorrectly. 🤦

PR #363, opened 2022-10-19
python: Fix a bug where transaction offsets weren't being set properly on reconnect

da-tanabe:
python: Fix a bug where transaction offsets weren't being set properly on reconnect.


alexgraham-da (approved):
Time is an illusion

PR #365, opened 2022-11-09
python: Merge v7 fixes into main

da-tanabe:
Backports of #362 and #363 into main:

- python: Add a timeout parameter to method calls to allow for specifying per-call timeouts, and for establishing Connection-scoped timeouts.
- python: Fix a bug where transaction offsets weren't being set properly on reconnect.


alexgraham-da (approved):
LGTM

PR #366, opened 2022-11-11
meta: Update the Daml version to pick up new gRPC services and Protobuf definitions in Daml 2.5.

da-tanabe:
meta: Update the Daml version to pick up new gRPC services and Protobuf definitions in Daml 2.5.

Note that none of these new APIs are exposed through dazl's APIs yet; this is merely updating gRPC stubs in accordance with the newest snapshot version of Daml 2.5.


da-tanabe:
Obsoleted by #377.

PR #367, opened 2022-11-12
docs: Remove references to dazl.model, as that module has been removed.

da-tanabe:
docs: Remove references to `dazl.model`, as that module has been removed in v8.


alexgraham-da (approved):
👋 file

PR #368, opened 2022-11-12
python: Remove dazl.util.tools.get_matches.

da-tanabe:
python: Remove dazl.util.tools.get_matches; it's been deprecated for a while and not used internally in dazl any more.


alexgraham-da (approved):
👋

PR #369, opened 2022-11-12
python: Remove the deprecated pathify function.

da-tanabe:
Remove `pathify`; it duplicates functionality in the Python standard library and is no longer used internally in dazl.


alexgraham-da (approved):
👋 function

PR #370, opened 2022-11-12
python: Remove the deprecated setup_default_logger function.

da-tanabe:
python: Remove the deprecated `setup_default_logger` function.

`dazl` library users should instead prefer to set up their own logger.

This code suffices as a replacement for most use cases:

```python
import logging
logging.basicConfig()
```


alexgraham-da:
Is the logger requirement in the docs?

da-tanabe:
> Is the logger requirement in the docs?

It's more that it's weird for a library to bring its own logger… but you're right, I'll add something to the migration guide.

PR #375, opened 2022-11-28
python: Add tests for metering in Daml 2.3.

da-tanabe:
python: Add tests that target the metering endpoints in Daml 2.3.

This is ahead of getting them to work with Daml 2.4, which changed the way that these services worked.


Sarahb-da (approved):
Nice! Pending test passes this lgtm

PR #376, opened 2022-11-28
python: Upgrade to Protobufs as they are defined in Daml 2.4.

da-tanabe:
Upgrade to using Daml 2.4's protobuf definitions.

Note that because the admin metering service changed in a backwards-incompatible way from 2.3 to 2.4, this means these bindings, when used specifically for admin metering service calls, will now only work with Daml 2.4 and later. This does _not_ affect compatibility for any other API, and the rest of the bindings remain backwards compatible to Daml 1.0.


alexgraham-da (approved):
Nice 👍

PR #377, opened 2022-11-29
python: Upgrade to Daml 2.5 protobufs (pre-release).

da-tanabe:
python: Upgrade to Daml 2.5 protobufs from the latest snapshot.

As the Daml 2.5 release is coming up soon, the APIs are unlikely to change at this point, so re-generate client stubs according to the latest Ledger API and Daml-LF 1.15 definitions.

This also switches the parser to using Daml-LF 1.15 definitions, which are still backwards compatible.


alexgraham-da (approved):
👍

PR #381, opened 2022-12-08
python: Support for DARs that contain interfaces.

da-tanabe:
python: Add support for DARs that contain interfaces.

Note that this doesn't add _full_ support for interfaces. If you use these bindings against a DAR that contain an interface, it will simply ignore them instead of failing to parse DARs.


_fixtures/src/kitchen-sink-sdk24/daml.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2017-2022, Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+sdk-version: 2.4.0
+name: kitchen-sink-sdk24
+version: 1.0.0
+build-options: [--target=1.dev]
```
dasormeter, line 7: do we want to target dev here or 1.15 explicitly?

python/tests/unit/test_ledger_submit.py
```diff
@@ -15,7 +15,7 @@ async def test_command_submission_with_stdlib_values(sandbox: SandboxLauncher) -
 
     async with dazl.connect(url=sandbox.url, admin=True) as conn:
         party_info, _ = await gather(
-            conn.allocate_party(), conn.upload_package(dars.KitchenSink.read_bytes())
+            conn.allocate_party(), conn.upload_package(dars.KitchenSink1_18.read_bytes())
```
dasormeter, line 18: why specific to only the 1_18 version?

scripts/make/dars
```diff
@@ -21,7 +21,7 @@ do
   case "${1}" in
   -M)
     echo "${dar}: ${daml_yaml} $(find "${daml_dir}" -name '*.daml' -type f | tr '\n' ' ')"
-    echo "	cd \"${daml_dir}\" && DAML_SDK_VERSION=1.18.1 daml build"
+    echo "	cd \"${daml_dir}\" && daml build"
```
dasormeter, line 24: why drop version?  this only builds latest 2.5.0--snapshot?  what about the 1.18.1 alternative?

.circleci/install-daml
```diff
@@ -8,7 +8,7 @@ function install {
         curl -sSL https://get.daml.com/ | sh
         link
     fi
-    daml install 2.4.0
+    daml install 2.5.0-snapshot.20221201.11065.0.caac1d10
```
dasormeter, line 11: do we need to install 2.4.0 as well still?

python/tests/unit/test_ledger_submit.py
da-tanabe: I wanted to clearly delineate the version of the library used, but not necessarily move all tests that use this model to test against all three variants.

scripts/make/dars
da-tanabe: This overrides the value in the `daml.yaml` file. I'm not sure why I originally did it this way, because I spent a fair amount of time trying to figure out why my versions in `daml.yaml` weren't taking effect… :-)

.circleci/install-daml
da-tanabe: hmm actually we probably do, since some of the DARs still build against 2.4.0—will add back.

_fixtures/src/kitchen-sink-sdk24/daml.yaml
da-tanabe: I'll change it. I was originally thrown off by the error message I was getting because… this was compiling under 1.18.1 instead of 2.4.0 (see my reply to your other comment below on removing `DAML_SDK_VERSION`).

python/tests/unit/test_ledger_submit.py
da-tanabe: (so to be clear, this is _not_ a change in behavior in these tests—it remains testing against the same model as before, and I just renamed it for clarity)

alexgraham-da (approved):
LGTM

PR #385, opened 2022-12-20
python: Better support for interfaces.

da-tanabe:
python: Fix some problems introduced by DARs with interfaces.

I'm not sure why the test cases introduced in #384 didn't catch these problems, because it really seems like they should have. So far parsing errors in interfaces have been reproduced by uploading them to a running ledger, and pulling them down through using the older `Network` client's synchronous package management handling scheme. I still haven't managed to reproduce this using the newer and simpler API.

I'll continue to take a look at this, but opting on a releasing a new version that I've manually tested ahead of more investigation as to why this can't seem to be repro'ed in a smaller and more straightforward test.


alexgraham-da (approved):
👍

PR #390, opened 2022-12-24
meta: Bump Daml SDK versions to 1.18.3 and 2.5.0.

da-tanabe:
Bump versions to Daml 1.18.3 and 2.5.0.

This also drops a test case that built a DAR using 2.4.0 to avoid taking a runtime dependency on a third version of the Daml SDK. The test cases added in #384 ensure that dazl continues to run properly against DARs built against 2.4.


dasormeter (approved):
LGTM

PR #404, opened 2023-05-09
python: Bump up dependencies, notably Sphinx.

da-tanabe:
Trying to modernize the `dazl` codebase a bit more.

When `dazl` was first started 5+ years ago, Sphinx 4 was a good idea… and we never upgraded it since. So bump up the versions, and make minor tweaks to the Sphinx setup so things continue to run properly.


dasormeter (approved):
🍕

.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         type: string
     docker:
       - image: cimg/python:<<parameters.python_version>>-browsers
-    resource_class: medium+
+    resource_class: large
```
dasormeter, line 64: intentional change?

PR #416, opened 2023-09-30
python: Drop quotes around types.

da-tanabe:
Quotes around types are no longer needed with `from __future__ import annotations`, and this has been available in Python for a while. Now that `main` on `dazl`'s minimum supported Python version is Python 3.8, drop these quotes.

There were a few places where I also added `/` to formally denote functions that take a single unnamed parameter, also a Python 3.8-ism.


dasormeter (approved):
so

PR #426, opened 2023-10-26
Add `[build-system]` to `pyproject.toml`

Ericson2314:
I think this is standard so tooling knows what to get to interpret the `[tool.poetry.*]` sections?

In any event, I needed this information in order to package and build this project (from the PyPI sdist, not git repo) in Nixpkgs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,3 +80,7 @@ log_cli = true
 log_cli_level = "DEBUG"
 junit_family = "xunit1"
 testpaths = ["python/tests/structure", "python/tests/unit"]
+
+[build-system]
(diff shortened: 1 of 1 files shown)


da-tanabe:
Sorry, I didn't notice this PR until now!

You're right—this is something that should be added. `dazl` has been using Poetry for a long time, going back to some prerelease versions that may not have included these by default.

We require a Contributor License Agreement to be signed to accept external contributions, but given the nature of this particular change, I also remade it as #428 to expedite merging in the change.

da-tanabe:
Merged in #428 in favor of this PR.

Ericson2314:
Thank you!

PR #429, opened 2023-11-10
python: Drop the custom .pyi plugin when generating Protobuf code.

da-tanabe:
I added support for `.pyi` files for Protobuf code via a custom `protoc` plugin (#271) back in August 2021. Unfortunately, just seven short months later, `protoc` added support for generating Python stubs with `--pyi_out` in [Protocol Buffers v3.20.0](https://github.com/protocolbuffers/protobuf/releases/tag/v3.20.0).

I just noticed it's a thing now, so switch to it instead of continuing to rely on the hand-rolled one I had to write. Unfortunately stubs are only generated for Protobuf messages—I still need to generate typing files for gRPC stubs myself for now.

There are some annoying things that mypy doesn't like about Protobuf's `.pyi` files, so post-process the `.pyi` files out of `protoc` and add `# type: ignore`. It might be a better solution long-term to tell mypy to ignore all of the generated protoc code if/when gRPC stubs also gain the ability to have typing files generated for them natively.


mschaef-da (approved):
Nice improvement.

PR #445, opened 2024-01-30
python: Redefine Command as a sealed type.

da-tanabe:
This is a bit of an experiment on the ergonomics of defining:

```python
Command = CreateCommand | ExerciseCommand | CreateAndExerciseCommand | ExerciseByKeyCommand
```

instead of the more traditional:

```python
class Command: ...
class CreateCommand(Command): ...
class ExerciseCommand(Command): ...
class CreateAndExerciseCommand(Command): ...
class ExerciseByKeyCommand(Command): ...
```

The advantage of using union types is mypy and other Python typecheckers can do a much better job of exhaustiveness checks. The disadvantage is that under Python 3.8 and Python 3.9, using union types breaks `isinstance` checks. This PR hacks around it by using defining alternate implementations for those two versions of Python, but keeping the typing definition union-based. (Side note: unfortunately Python 3.9 is scheduled for EOL in October 2025 so we'll be stuck with this awkwardness for a while.)

PEP 622, the original proposal for structural pattern matching included [sealed types](https://peps.python.org/pep-0622/#sealed-classes-as-algebraic-data-types), but [PEP 634](https://peps.python.org/pep-0634/), which succeeded it, dropped them. That would allowed us to use the best of both options, but sadly it isn't an option.

### DIFF
--- a/python/tests/unit/test_command.py
+++ b/python/tests/unit/test_command.py
@@ -1,0 +1,22 @@
+# Copyright (c) 2017-2024 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from dazl import Command, ContractId, CreateCommand
+from dazl.damlast.lookup import parse_type_con_name
+
+template_id_str = "deadbeef:DeadBeef:DeadBeef"
+template_id = parse_type_con_name(template_id_str)
+cid = ContractId(template_id, "#0:0")
+choice = "Cook"
+payload = {"operator": "beef", "weight": "10kg"}
+key = {"operator": "beef"}
+
+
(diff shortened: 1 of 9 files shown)


python/tests/unit/test_command.py
```diff
@@ -0,0 +1,80 @@
...
+                case ExerciseByKeyCommand(
+                    matched_template_id, matched_key, matched_choice, matched_arg
+                ):
+                    assert template_id == matched_template_id
+                    assert key == matched_key
+                    assert choice == matched_choice
+                    assert matched_arg == {}
+                    found_exercise_by_key = True
+                case _:
+                    assert_never(cmd)
```
da-tanabe, line 75: This is the main dish: full exhaustiveness that typechecks with Python 3.10's `match` statement.

PR #453, opened 2024-04-04
security policy for dazl client

dasormeter:
Define a security policy for dazl-client (similar to the one for daml)


dasormeter:
@nycnewman / @dtanabe : does this security policy look reasonable for dazl client?

nycnewman (approved):
LGTM

PR #457, opened 2024-04-13
python: Drop support for Python 3.8 and 3.9.

da-tanabe:
It's too difficult to properly support Python 3.10's `match` feature, as well as proper support for union types (`isinstance` checks), while also retaining backwards compatibility support for Python 3.8 and Python 3.9.

Python 3.8 will be end-of-lifed in six months (October 2024), but Python 3.9 will still be around for another year (October 2025). Nevertheless, dropping support for these versions of Python will make maintaining this code much easier (and hopefully also shrink it some too).


dasormeter (approved):

da-tanabe:
I spent some time looking for stats on what versions of Python are actually in use, and finally found something suitable to base a decision on: https://pypistats.org/packages/__all__

This unfortunately shows that doing this _now_ is probably too early… 3.10+ doesn't have a lot of market share, and even scarier, lots of people are even using EOL'ed versions of Python.

A snapshot taken on April 18, 2024:

| Python | "total"* |
| -- | --: |
| ≥3.6 | 98.85% |
| ≥3.7 | 94.36% |
| ≥3.8 | 78.05% |
| ≥3.9 | 59.90% |
| ≥3.10 | 40.45% |
| ≥3.11 | 16.70% |
| ≥3.12 | 4.06% |
| ≥3.13 | 0.01% |

So going from supporting Python 3.8+ to Python 3.10+ would drop the user base from 78.05% to 40.05%. As much I'd like to take advantage of newer Python-isms, sadly it seems like the community isn't ready for that yet. :-/

PR #472, opened 2024-07-11
python: Drop the old ledger client.

da-tanabe:
python: Remove the old `dazl.client` package and its supporting infrastructure. This code still exists on the v7 branch, where it will continue to be maintained for as long as that is necessary.


dasormeter (approved):

alexgraham-da (approved):
🔥 🔥 🔥

PR #483, opened 2024-12-10
Correct several scenarios related to using user tokens against a local sandbox instance

mschaef-da:


pyproject.toml
```diff
@@ -16,7 +16,7 @@ packages = [{ include = "dazl", from = "python" }, { include = "_dazl", from = "
 [tool.poetry.dependencies]
 python = "^3.8"
 googleapis_common_protos = "^1"
-grpcio = ">=1.65.0"
+grpcio = "1.67.1"
```
mschaef-da, line 19: Going to the latest `1.68.1` breaks reconnect tests that hit the ledger due to the fact (seemingly) that the ledger API is still implemented on `1.67.1`

See: https://github.com/grpc/grpc/issues/31442

da-tanabe (approved):
Thanks!

PR #486, opened 2024-12-12
go: Fix some invalidly-generated code.

da-tanabe:
Make the `go` code for `dazl` actually behave a bit better.

Upgrade the `grpc` go plugins, as well as bump the minimum required Go version to 1.22, mostly to keep up with upstream go gRPC libraries, which no longer compile on older versions.

Also move from v7 to v8 in the module name, as `main` represents a beta version of v8.


jordan-da (approved):
thank you very much

PR #503, opened 2025-04-11
python: Drop Daml v1 support.

da-tanabe:
python: Drop support for Daml v1. The next version of dazl will no longer work against Daml v1 ledgers.

In particular, dazl will no longer attempt to fetch a `ledger_id` from the remote if one was not already supplied at connection construction time. If a user knew their ledger ID and supplied it, there is a pretty good chance that dazl could continue to work in that configuration against Daml v1 ledgers.

Just as points of reference:
* Daml v1.18.3, the last version of the "1" line, came out in June 2022.
* Daml v2.0.0 was released in February 2022.

The tests are slightly faster as an extra Daml v1 sandbox is no longer spun up and those extra tests are no longer run.

This PR also upgrades from 2.9.6 to 2.10.0.


dasormeter (approved):

da-tanabe:
> @da-tanabe will this affect the hub models that are still on daml 1.x? Or only if you are running a 1.x ledger version for the sandbox/participant?

Yup correct, this doesn't affect DAR compatibility at all (and those tests are still in place, too)—this is just about connecting and running against a Daml 1.x ledger.

PR #516, opened 2025-05-14
use nix for pip and setuptools

dasormeter:


da-tanabe:
hmm these should be getting pulled in by the `poetry` instead...

PR #518, opened 2025-05-19
upgrade urllib3

dasormeter:


pyproject.toml
```diff
@@ -22,6 +22,7 @@ pygments = { version = "*", optional = true }
 pyOpenSSL = { version = "*", optional = true }
 semver = "*"
 typing_extensions = { version = "*" }
+urllib3 = "2.4.0"
```
da-tanabe, line 25: hmm can you try just running `poetry update` instead of adding a direct pinned dependency?

This is problematic for two reasons:
* dazl doesn't have a direct dependency on `urllib3`, so this is inaccurate
* You also specified an explicit version (2.4.0), which would prevent people from being able to upgrade in the future as further patch releases may be needed.

dasormeter: okay - updated now based on `poetry update` and that also did tick urllib3 to 2.4.0

da-tanabe (approved):

PR #524, opened 2025-06-12
meta: Upgrade Daml to 2.10.1 and Canton protos to 3.3.0.

da-tanabe:
Update protobuf bindings.

Note that the Python bindings were actually generated with `grpcio-tools` 1.72.1 instead of 1.73.0 because of https://github.com/protocolbuffers/protobuf/issues/22178.


jordan-da (approved):
this crashed my browser

PR #576, opened 2026-04-23
uptick openssl

dasormeter:
uptick openssl to address sec vuln


pyproject.toml
```diff
@@ -21,7 +21,7 @@ grpcio = "^1.72.0"
 httpx = "^0.27.0"
 protobuf = ">=6"
 pygments = { version = "*", optional = true }
-pyOpenSSL = { version = "*", optional = true }
+pyOpenSSL = "26.0.0"
```
da-tanabe, line 24: hrmm why did you drop `optional`? And by pinning it to a specific version, this prevents users from ever updating this on their own as well…

PR #589, opened 2026-06-10
[bug] fixing import cycle in Golang

moldis:
Latest proto update introduced cycle import issue in Go, can be reproduced only if import all photo files in project (which is go-daml doing).

* The GlobalKeyWithMaintainers message was moved into interactive_submission_common_data.proto.
* That proto file declares proto package com.daml.ledger.api.v2.interactive — the same package as interactive_submission_service.proto. The generator maps proto-package → Go-package 1:1, so both land in the same Go package …/interactive.
* That creates a two-way edge between two Go packages:
- …/interactive (the service) imports …/interactive/transaction/v1 — it needs v1.Create, v1.Node.
- …/interactive/transaction/v1 imports …/interactive — it needs GlobalKeyWithMaintainers, which now lives there.

My fix is moving GlobalKeyWithMaintainers to common package and updating CI/CD to build whole protos (to avoid cycles in future builds)

The error:
```
imports github.com/digital-asset/dazl-client/v8/go/api/com/daml/ledger/api/v2/interactive from converters.go
        imports github.com/digital-asset/dazl-client/v8/go/api/com/daml/ledger/api/v2/interactive/transaction/v1 from interactive_submission_service.pb.go
        imports github.com/digital-asset/dazl-client/v8/go/api/com/daml/ledger/api/v2/interactive from interactive_submission_data.pb.go: import cycle not allowed
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,6 +88,26 @@ jobs:
           command: make python-typecheck
       - store_test_results:
           path: target/test-results
+  check-go:
+    docker:
+      - image: cimg/base:stable
+    resource_class: medium
+    steps:
+      - checkout
+      - run:
+          name: Nix preparation
+          command: .circleci/scripts/prep.sh
+      - restore_cache:
+          name: Nix restore cache
+          keys:
+            - nix-{{ checksum "/home/circleci/nix/cache-keys/nix-checksums" }}
+            - nix
+      - run:
+          name: Nix Setup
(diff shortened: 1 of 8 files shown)


da-tanabe (approved):

da-tanabe:
@moldis can you rebase your branch against `main`? thanks!

moldis:
> @moldis can you rebase your branch against `main`? thanks!

Done!

